Issue #29, opened 2021-06-01
Question: Different I/O for different methods


glitch452:
Hi,

I've started trying to use this module to manage my endpoints since I love the philosophy of how it's setup.  I have a question about a use case.  

For an endpoint with multiple methods (i.e. `post` and `get`) is it possible to have an input and output for `post` that is different than the input and output for `get`? Or, I guess a way to attach different endpoint definitions to the same route would also work?

RobinTail:
Hello @glitch452 ,

Thank you for the question. 

Essentially, an endpoint currently can hold the only input and output schema. And a routing path can be attached to the only endpoint.

I made `Endpoint` possible to handle different methods, but my idea here was to handle [RESTful](https://searchapparchitecture.techtarget.com/definition/RESTful-API) requests with similar parameters: 

- GET and DELETE: require an entity id to obtain or remove it;
- POST and PUT: require an entity properties to be create or update it.

The variable behavior of the handler depending on the actual `Method` can be achieved by using `methodProviderMiddleware` ([in example](https://github.com/RobinTail/express-zod-api/blob/6c4a46f5e193446beb5b2286ba6de10afad393ac/example/middlewares.ts#L19)). 

However, the endpoint I/O schema is still the only one, therefore I can offer you two solutions for your objective.

1. Use different paths and different endpoints for different I/O schemas.

```typescript
export const routing: Routing =  {
  v1: {
    user: {
      get: yourEndpoint1, // method: GET, input schema with id, output schema with properties
      create: yourEndpoint2, // method: POST, input schema with properties, output schema with new id
    }
  }
};
```

So you will have the following independent paths like this:

- https://example.com/v1/user/get
- https://example.com/v1/user/create

2. Use `.or()` next to `z.object()`.

You can specify multiple I/O object schemas using `.or()` concatenation and then just check the presence of some uniq parameter using `in` condition, so that the compiler understands which particular schema is used for a given condition.

Here is an example:

```typescript
export const universalEndpoint = endpointsFactory
  .addMiddleware(methodProviderMiddleware)
  .build({
    methods: ['get', 'post'],
    input: z.object({ // for GET
      id: z.string()
    }).or(z.object({ // for POST
      name: z.string(),
      age: z.number(),
      address: z.string()
    })),
    output: z.object({ // for GET
      name: z.string(),
      age: z.number(),
      address: z.string()
    }).or(z.object({ // for POST
      id: z.number()
    })),
[...]

PR #32, opened 2021-06-05
More flexible methods for Endpoint and Routing

RobinTail:
1. A more convenient way to specify a single method to the Endpoint.
2. Ability to assign multiple Endpoints to a single Route depending on Method (the issue: https://github.com/RobinTail/express-zod-api/issues/29)
3. Ability for the Endpoint to handle multiple methods of the single Route at the same time


src/method.ts
```diff
@@ -0,0 +1,41 @@
+import {Endpoint} from './endpoint';
+
+export type Method = 'get' | 'post' | 'put' | 'delete' | 'patch';
+
+export type MethodsDefinition<M extends Method> = {
+  methods: M[];
+} | {
+  method: M;
+};
```
RobinTail, line 9: Two ways to specify the Endpoint method(s).

example/v1/index.ts
```diff
@@ -1,8 +1,13 @@
...
 
-export const v1Routing: Routing =  {
+export const v1Routing: Routing = {
+  // syntax 1: methods are defined within the endpoint
   getUser: getUserEndpoint,
-  setUser: setUserEndpoint
+
+  // syntax 2: methods are defined within the route
+  setUser: new RouteMethods({
+    post: setUserEndpoint // the Endpoint should have at least the same method specified in .build()
```
RobinTail, line 11: A route can have multiple handlers (with different I/O schemas) depending on different methods.
It can also be the same Endpoint that handle different methods as well.

example/v1/index.ts
```diff
@@ -1,8 +1,13 @@
...
 
-export const v1Routing: Routing =  {
+export const v1Routing: Routing = {
+  // syntax 1: methods are defined within the endpoint
   getUser: getUserEndpoint,
-  setUser: setUserEndpoint
+
+  // syntax 2: methods are defined within the route
+  setUser: new DependsOnMethod({
+    post: setUserEndpoint // the Endpoint should have at least the same method specified in .build()
```
RobinTail, line 11: A route can have multiple handlers (with different I/O schemas) depending on different methods.
It can also be the same Endpoint that handle different methods as well.

PR #48, opened 2021-06-15
Fix possible CORS issue with OPTIONS pre-flight request.

RobinTail:


tests/system/example.spec.ts
```diff
@@ -42,6 +42,20 @@ describe('Example', () => {
       })).toBeTruthy();
     });
 
+    test('Should handle OPTIONS request', async () => {
+      const response = await fetch('http://localhost:8090/v1/setUser?test=123&id=50', {
+        method: 'OPTIONS',
+      });
+      expect(response.status).toBe(200);
```
RobinTail, line 49: should also check the empty response body?

PR #49, opened 2021-06-18
V2: Disclosing the complete response of endpoints

RobinTail:


src/endpoints-factory.ts
```diff
@@ -13,29 +13,26 @@ type BuildProps<IN extends IOSchema, OUT extends IOSchema, mIN, mOUT, M extends
...
-    factory.resultHandler = resultHandler;
     return factory;
   }
 
-  public setResultHandler(resultHandler: ResultHandler) {
-    return EndpointsFactory.create<mIN, mOUT>(
-      this.middlewares,
-      resultHandler
-    );
-  }
```
RobinTail, line 35: Breaking. Use `new EnpointsFactory()` instead

PR #97, opened 2021-08-12
Upgrade mime dependency from v1 to v2

nathanpotter:
First off I wanted to say thanks for putting together this project, I really enjoy how you designed things!

I'm currently trying to use this within a monorepo as a part of a yarn workspace, but several of our other dependencies are relying on mime v2 instead of mime v1 which is causing some issues. This PR updates the mime dependency within express-zod-api.

Let me know if you'd like to go another direction with it and I can adjust things :)

### DIFF
--- a/example/factories.ts
+++ b/example/factories.ts
@@ -1,4 +1,4 @@
-import {lookup} from 'mime';
+import {getType} from 'mime';
 import {
   createApiResponse,
   createResultHandler,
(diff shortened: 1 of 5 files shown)


RobinTail:
Hello @nathanpotter .
Thank you for the feedback.

> mime v1 which is causing some issues

I'm sorry, I do not understand. Could you please clarify how the dependency within `express-zod-api` affects your repo?
What is the issue?
How it prevents you from using `mime@v2` within _your_ repo, because `yarn` should be capable for building isolated dependency graph in `node_modules` dir.

Besides that your PR does *not* remove the `mime@v1` — it adds `mime@v2` along with it:

image

FYI, `mime@v1` is the dependency of the latest `express`, so it can not be removed.

```
$ yarn why mime
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "mime"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "mime@1.6.0"
info Reasons this module exists
   - "express#send" depends on it
   - Hoisted from "express#send#mime"
```

Looking forward to your reply.

nathanpotter:
Apologies, you're correct - I believe I have something misconfigured in my repo. I've also decided to go another route as well that isn't zod based so I unfortunately won't be able to investigate this issue further. I'm going to close this PR, but thank you for your thoughtful reply.

Issue #98, opened 2021-08-14
Zod type ZodNonEmptyArray is unsupported by OpenAPI


kirdk:
If output uses `z.array().notempty`, Open API won't start with the message `Zod type ZodNonEmptyArray is unsupported`
`ZodUndefined` also does not work
express-zod-api version is `2.3.0`

RobinTail:
Hello @kirdk .

Thank you for the feedback.

> `z.array().notempty`

I presume you mean `z.array(...).nonempty()`. In this case I can not reproduce the issue in `v2.3.0`, it works with a following result
```yaml
schema:
  type: array
  items:
    type: ...
  minItems: 1
```

Could you please provide more context or code sample for me to reproduce the issue.


> `ZodUndefined` also does not work

I didn't expect this to be used given that there is `.optional()` refinement, but I'll add it, no problem.

Thank you

kirdk:
D

> Hello @kirdk .
> 
> Thank you for the feedback.
> 
> > `z.array().notempty`
> 
> I presume you mean `z.array(...).nonempty()`. In this case I can not reproduce the issue in `v2.3.0`, it works with a following result
> 
> ```yaml
> schema:
>   type: array
>   items:
>     type: ...
>   minItems: 1
> ```
> 
> Could you please provide more context or code sample for me to reproduce the issue.
> 
> > `ZodUndefined` also does not work
> 
> I didn't expect this to be used given that there is `.optional()` refinement, but I'll add it, no problem.
> 
> Thank you

Dear @RobinTail, I have checked again with v. 2.3.1 and the issue is the same. The problem appears when the response schema contains a ZodNonEmptyArray.
![image](https://user-images.githubusercontent.com/12613557/129705449-b01935ce-908a-4841-b66b-4a5a1083633a.png)

Regarding ZodUndefined. In some cases, it's useful to implement subTypes. Because if we define the subtypes something like:
`
const Base = z.object({ id: z.string() });
const SubType1 = Base.extend({ field1: z.string() });
const SubType2 = Base.extend({ field2: z.string() });
const Type = z.union([SubType1, SubType2]);
type Type = z.infer<typeof Type>;
`
We can have a compilation error:
![image](https://user-images.githubusercontent.com/12613557/129706295-d0af6254-65cf-46ac-aa76-f61db0230952.png)
But if we define the schema like:
`
const Base = z.object({ id: z.string() });
[...]

PR #99, opened 2021-08-14
Supporting type ZodUndefined in OpenAPI/Swagger.

RobinTail:
Related to the issue #98


RobinTail:
This will not be merged since
1. There is an `.optional()` refinement.
2. There is no need for this, since the original issue can be fixed in a more simple way: #104

PR #112, opened 2021-08-22
Draft: Better snapshots

RobinTail:


tests/system/__snapshots__/system.spec.ts.snap
```diff
@@ -0,0 +1,92 @@
...
+Object {
+  "error": Object {
+    "fields": Object {
+      "anything": Array [
+        "Value should be greater than 0",
+      ],
+      "output": Array [
+        "Invalid format",
+      ],
+    },
```
RobinTail, line 59: I'm not happy with it.
It seems that `.flatten()` takes only the first item from the path for making fields in `fieldErrors`.

tests/unit/__snapshots__/endpoint.spec.ts.snap
```diff
@@ -0,0 +1,51 @@
...
+          },
+        },
+        "message": Object {
+          "_type": "ZodString",
+        },
+      },
+    },
+    "status": Object {
+      "_type": "ZodLiteral",
+    },
```
RobinTail, line 29: need to disclose the value of literal schema in helper

RobinTail: fixed in 6ccd509780c0d0663dabd006d71af3b3aa7682ea

tests/system/__snapshots__/system.spec.ts.snap
RobinTail: fixed in 90a711264a06ff6a8bf38b5cba3187def2ca7323

RobinTail:
- I wanna have a configuration option for this feature
- I wanna make it available for a customized ResultHandler

PR #201, opened 2021-11-25
Fixing OpenAPI path params format

punctsivirgula:
Hi Anna,

At least according to [this issue](https://github.com/OAI/OpenAPI-Specification/issues/1681), 'colon-style' path parameters (`/entity/:id`, as used by express) are not supported by the OpenAPI specification; instead the 'curly braces notation' (`/entity/{id}`) is preferred. The behaviour is deemed as 'tooling issue' and, as such, it is up to the tool generating the OpenAPI document to adjust it's output to match the specification. I'm no expert in OpenAPI related things, so my interpretation might be wrong. Nevertheless, I've tried to fix the issue. Hope I've done a not too terrible job, but I'm open to corrections.

I've just discovered your library today, after quite a while of searching for something similar and giving up at least half a dozen times after not finding enough time to develop something myself. Since I only had a few hours of testing the library I'm still preparing myself for future issues, but so far I love it. Great work!

### DIFF
--- a/src/open-api.ts
+++ b/src/open-api.ts
@@ -7,6 +7,14 @@ import {
 } from "./open-api-helpers";
 import { Routing, routingCycle, RoutingCycleParams } from "./routing";
 
+const convertColonToBrackets = (path: string): string =>
+  path
+    .split('/')
+    .map(segment =>
+        segment[0] === ':' ? `{${segment.substr(1)}}` : segment,
+    )
+    .join('/')
+
 interface GeneratorParams {
   title: string;
   version: string;
@@ -55,6 +63,7 @@ export class OpenAPI extends OpenApiBuilder {
         // @todo involve config/inputSources in v4
         operation.requestBody = depictRequest(commonParams);
       }
+      path = convertColonToBrackets(path);
       this.addPath(path, {
         ...(this.rootDoc.paths?.[path] || {}),
         [method]: operation,


RobinTail:
Wow, thank you so much @punctsivirgula .
I did not know about such issue, and the Swagger validation also passed well.
I need some time to investigate it.

Regarding the suggested changes.
In general — I see your idea and I believe it might work great in most cases,
However, I'm going to suggest a couple of corrections to your implementation in order to cover some edge cases and make it more reliable.

I'll get back to it soon. Thank you

RobinTail:
ESLint error in CI: need to run `yarn eslint ./src --fix`

src/open-api.ts
```diff
@@ -55,6 +63,7 @@ export class OpenAPI extends OpenApiBuilder {
         // @todo involve config/inputSources in v4
         operation.requestBody = depictRequest(commonParams);
       }
+      path = convertColonToBrackets(path);
       this.addPath(path, {
```
RobinTail, line 67: I'm not a big fan of mutable stuff, so I try to avoid it where possible.
Luckily, here it's easy:

```typescript
const swaggerSafePath = convertColonToBrackets(path);
this.addPath(swaggerSafePath, {
  ...(this.rootDoc.paths?.[swaggerSafePath] || {}),
  [method]: operation,
});
```

src/open-api.ts
```diff
@@ -7,6 +7,14 @@ import {
 } from "./open-api-helpers";
 import { Routing, routingCycle, RoutingCycleParams } from "./routing";
 
+const convertColonToBrackets = (path: string): string =>
+  path
+    .split('/')
+    .map(segment =>
+        segment[0] === ':' ? `{${segment.substr(1)}}` : segment,
```
RobinTail, line 14: It's possible that path parameter is used not in the beginning of the segment.
Example: `/flights/:from-:to` from https://expressjs.com/en/guide/routing.html.

That's why I'm using a regular expression recommended on the same page in order to obtain the path params.
Consider using `String.replace(//)` with the same RegExp I use in the [common helper getRoutePathParams()](https://github.com/RobinTail/express-zod-api/blob/c3f77c77a54eb87513a377c7c23adeeebc9954b2/src/common-helpers.ts#L220).

RobinTail (changes requested):
I'm requesting two changes.
+ please run linter in order to avoid errors in CI.
Good job!

RobinTail:
@punctsivirgula ,

Since I wanted to fix the issue as soon as possible, I made my own PR #203 based on your suggestions.

Therefore, I am closing this request, but I want to express my gratitude to you for reporting the issue and your contributions to the open source community.

Thank you so much @punctsivirgula and your future contributions are appreciated.

RobinTail:
The fix has been released in version 3.1.2 🚀

PR #208, opened 2021-11-27
Breaking: Using config within OpenAPI generator

RobinTail:
Starting version 2.8.0 `inputSources` can be configured.
This should be taken into account in OpenAPI generator.

### DIFF
--- a/src/endpoints-factory.ts
+++ b/src/endpoints-factory.ts
@@ -21,8 +21,6 @@ type BuildProps<
   output: OUT;
   handler: Handler<z.output<Merge<IN, MwIN>>, z.input<OUT>, MwOUT>;
   description?: string;
-  /** @deprecated the factory automatically detects the usage of z.upload() within the input schema */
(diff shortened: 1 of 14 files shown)


src/endpoints-factory.ts
```diff
@@ -21,8 +21,6 @@ type BuildProps<
   output: OUT;
   handler: Handler<z.output<Merge<IN, MwIN>>, z.input<OUT>, MwOUT>;
   description?: string;
-  /** @deprecated the factory automatically detects the usage of z.upload() within the input schema */
-  type?: "json" | "upload"; // @todo remove in v4
```
RobinTail, line 25: breaking changes

src/open-api.ts
```diff
@@ -13,6 +15,7 @@ interface GeneratorParams {
   version: string;
   serverUrl: string;
   routing: Routing;
+  config: CommonConfig;
```
RobinTail, line 18: breaking changes

PR #229, opened 2021-12-10
Handler test running

Isaac-Leonard:
Add a method that can be called to test endpoint handlers

### DIFF
--- a/src/endpoint.ts
+++ b/src/endpoint.ts
@@ -1,9 +1,9 @@
+import { once } from "events";
 import { Request, Response } from "express";
-import { Logger } from "winston";
+import { Writable } from "stream";
+import Winston, { Logger } from "winston";
 import { z } from "zod";
 import { ApiResponse } from "./api-response";
-import { CommonConfig } from "./config-type";
-import { ResultHandlerError } from "./errors";
 import {
   combineEndpointAndMiddlewareInputSchemas,
   getInitialInput,
@@ -12,6 +12,8 @@ import {
   OutputMarker,
   ReplaceMarkerInShape,
 } from "./common-helpers";
+import { CommonConfig } from "./config-type";
+import { ResultHandlerError } from "./errors";
 import { Method, MethodsDefinition } from "./method";
 import { MiddlewareDefinition } from "./middleware";
 import { lastResortHandler, ResultHandlerDefinition } from "./result-handler";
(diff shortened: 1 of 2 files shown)


RobinTail:
Thank you for your PR @Isaac-Leonard , your contribution and ideas are highly appreciated.
I realized that I need to clarify in the documentation on how to test endpoints.

Unfortunately I can not accept the offered solution due to the imbalance of efforts in relation to benefits.
The reason is that the offered method tests the output of the endpoint's handler, instead of its actual response, but more importantly — overlooking the error handling implemented in the ResultHandler that is part of the Endpoint's workflow.

The handler may throw an error, but how exactly it will affect on the Endpoint — decides the ResultHandler, which may be customized by design as well.

Therefore, testing handler is not enough. What should be tested is the public `execute()` method of Endpoint.
And there is no need for additional methods, since it is public.

So, how to test endpoints.
1. Mock `request`, `response` and `logger`
2. Invoke `.execute({request, response, config, logger})`
3. Assert expectations on `response` and/or `logger` methods.

Here is an example:
```typescript
const requestMock = {
  method: "POST",
  header: jest.fn(() => mimeJson),
  body: {...},
};
const responseMock: Record<string, jest.Mock> = {
  set: jest.fn().mockImplementation(() => responseMock),
  status: jest.fn().mockImplementation(() => responseMock),
  json: jest.fn().mockImplementation(() => responseMock),
};
const configMock = {
  cors: true, // or your actual config
};
await endpoint.execute({
  request: requestMock as unknown as Request,
  response: responseMock as unknown as Response,
  config: configMock as CommonConfig,
  logger: loggerMock,
});

expect(loggerMock.error).toBeCalledTimes(0);
expect(responseMock.status).toBeCalledWith(200);
expect(responseMock.json).toBeCalledWith({
  status: "success",
  data: {...},
});
```

That's it.
I hope you'll find this helpful, and I'm going increase the visibility of this information.

Thank you once again for your contribution!

Isaac-Leonard:
I understand that testing the public execute function is generally preferable however for unit testing I would like to be able to test the raw handle method as well without having to set up mocks for request and response objects.
I realise now however that a much simpler way then my testing function would likely just be to remove the protected modifier from the handler method on the endpoint class.
This also allows custom options to be passed for testing such as a mock database connection or fake auth object without having to mock the middleware handlers.

PR #252, opened 2021-12-27
Endpoint testing helper

RobinTail:
```typescript
const { responseMock, loggerMock } = await testEndpoint({
  endpoint,
  requestProps: {
    method: "POST",
    body: {...},
  },
});
expect(loggerMock.error).toBeCalledTimes(0);
expect(responseMock.status).toBeCalledWith(200);
expect(responseMock.json).toBeCalledWith({
  status: "success",
  data: {...},
});
```

### DIFF
--- a/src/mock.ts
+++ b/src/mock.ts
@@ -1,0 +1,90 @@
+import { Request, Response } from "express";
+import { Logger } from "winston";
+import { CommonConfig } from "./config-type";
+import { AbstractEndpoint } from "./endpoint";
+import { createHttpError } from "./index";
+import { mimeJson } from "./mime";
+
+interface TestEndpointProps<REQ, RES, LOG> {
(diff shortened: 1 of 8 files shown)


src/mock.ts
```diff
@@ -0,0 +1,90 @@
...
+    {
+      writableEnded: boolean;
+      statusCode: number;
+      statusMessage: string;
+    } & Record<"set" | "status" | "json" | "end", jest.Mock> &
+      (RES extends undefined ? {} : RES)
+  >{
+    writableEnded: false,
+    statusCode: 200,
+    statusMessage: createHttpError(200).message,
```
RobinTail, line 54: > http-errors deprecated non-error status code; use only 4xx or 5xx status codes

use `statuses` instead: https://www.npmjs.com/package/statuses

PR #254, opened 2021-12-30
Ability to serve static files.

RobinTail:
The feature requested in the discussion
https://github.com/RobinTail/express-zod-api/discussions/249

```typescript
import {Routing, ServeStatic} from "express-zod-api";

const routing: Routing = {
  // path /public serves static files from ./assets
  public: new ServeStatic(path.join(__dirname, "assets"), {
    dotfiles: "deny",
    index: false,
    redirect: false,
  }),
};
```


example/routing.ts
```diff
@@ -23,4 +24,10 @@ export const routing: Routing = {
       upload: uploadAvatarEndpoint,
     },
   },
+  // path /public serves static files from /example/assets
+  public: serveStatic(path.join(__dirname, "assets"), {
```
RobinTail, line 28: perhaps it would be better to make it a class in order to preserve the consistency with `new DependsOnMethod`

PR #263, opened 2022-01-03
Supporting Node 17

RobinTail:


RobinTail:
There is no `@tsconfig/node17` yet.

RobinTail:
Seems to be operational on Node 17.

PR #270, opened 2022-01-08
Issue 269: async refinements

### LINKED ISSUE #269: Async refinements
Hello,

First of all, thank you very much for your package!

I'm starting using it and I'm running into an issue. I have a piece of code that looks like this:
```ts
input: z.object({
        username: z.string().nonempty().max(100).refine(async username => {
            return await someAsyncFunction(username} == 0
       })
}),
```
I'm doing an async refinement, which should be allowed according to Zod docs. But I am getting this error:
`Async refinement encountered during synchronous parse operation. Use .parseAsync instead.`.

It looks like my `username` is not being parsed with parseAsync() but with parse(). Is the fact that we can't use async refinements on purpose?

RobinTail:
Closes #269

### DIFF
--- a/src/endpoint.ts
+++ b/src/endpoint.ts
@@ -196,9 +196,9 @@ export class Endpoint<
     response.set("Access-Control-Allow-Headers", "content-type");
   }
 
-  #parseOutput(output: any) {
+  async #parseOutput(output: any) {
     try {
-      return this.outputSchema.parse(output);
+      return await this.outputSchema.parseAsync(output);
     } catch (e) {
       if (e instanceof z.ZodError) {
         throw new z.ZodError([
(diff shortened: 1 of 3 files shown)


src/endpoint.ts
```diff
@@ -231,7 +231,7 @@ export class Endpoint<
     const options: any = {};
     let isStreamClosed = false;
     for (const def of this.middlewares) {
-      input = { ...input, ...def.input.parse(input) }; // middleware can transform the input types
+      input = { ...input, ...(await def.input.parseAsync(input)) }; // middleware can transform the input types
       Object.assign(
```
RobinTail, line 235: Can use Object.assign() here as well. Might look prettier.

RobinTail:
Should be fine.

PR #286, opened 2022-01-16
Reorganizing dependencies of types

RobinTail:
Express-Zod-API is a Typescript library intended to be used in Typescript environment.
Therefore it provides own types.
But some of these types do actually refer to other libraries.
For example: there are entities used within `createConfig()` exported method, that depend on types of `express-fileupload`, `compression` and so on.
This may complicate writing the correct configuration, so I'm going to fix it.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies


RobinTail:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44777#issuecomment-629660992

RobinTail:
Released as v5.4.1-beta1 🚀

RobinTail:
Tested 👍🏽

PR #297, opened 2022-01-22
Dates problem resolution

RobinTail:
JSON does not support Date.
Using `z.date()` in IO is confusional.


src/extend-zod.ts
```diff
@@ -1,6 +1,18 @@
+import { ZodDate } from "zod";
```
RobinTail, line 1: double import here

RobinTail: should be fine
https://stackoverflow.com/questions/48951687/override-an-export-from-another-library-es6

RobinTail:
✅ Approved for beta

RobinTail:
🚀 Released as 5.5.0-beta1

RobinTail:
Tested — operational
https://github.com/RobinTail/incaseofmydeath/commit/39c0af48b7fa97da0c80022983b083132ce0152d

PR #340, opened 2022-03-05
v6: Replacing Merge with ZodIntersection

RobinTail:
### Improvements

- better generated documentation of `ZodUnion`s on the top level
- ability to use `ZodDiscriminatedUnion` on the top level

### Breaking changes

- Type narrowing: You cannot use the `.strict()`, `.passthrough()` and its deprecated alias `.nonstrict()` methods in middleware object schemas. 
- Only `.strip()` is allowed in middlewares, which is actually default, so you should not use any of them at all.


example/example.swagger.yaml
```diff
@@ -147,39 +147,40 @@ paths:
           schema:
             description: POST /v1/user/:id parameter
             type: string
-          examples:
-            example1:
-              value: "12"
```
RobinTail, line 152: this should not be gone

example/example.swagger.yaml
```diff
@@ -147,39 +147,40 @@ paths:
...
+                    key: 1234-5678-90
+                  type: object
+                  properties:
+                    key:
+                      type: string
+                      minLength: 1
+                  required:
+                    - key
+                - example:
+                    id: "12"
```
RobinTail, line 166: URL param example should not be mixed in

example/example.swagger.yaml
RobinTail: fixed in 0aec7f2

example/example.swagger.yaml
RobinTail: fixed in c25577d

RobinTail:
Ready for experiment ✅

RobinTail:
Released as `5.9.0-beta1` via `beta` channel 🚀

RobinTail:
Operational ✅

src/middleware.ts
```diff
@@ -17,16 +17,22 @@ type Middleware<IN, OPT, OUT> = (
 ) => Promise<OUT>;
 
 export interface MiddlewareDefinition<
-  IN extends IOSchema,
+  IN extends IOSchema<"strip">,
```
RobinTail, line 20: breaking ⚠️

PR #362, opened 2022-03-20
Experiment: ResultHandler running TS 4.7

RobinTail:
`https://github.com/microsoft/TypeScript/pull/47607`


RobinTail:
The feature of TS 4.7 is unfortunately not enough.

PR #408, opened 2022-04-23
Breaking: Utility type removal

RobinTail:
This is for the next major release in case the client generator would prove its reliability.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,5 @@
 export { createConfig, LoggerConfig } from "./config-type";
-export {
-  AbstractEndpoint,
-  EndpointInput,
-  EndpointOutput,
-  EndpointResponse,
-} from "./endpoint";
+export { AbstractEndpoint } from "./endpoint";
 export { Method } from "./method";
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -1,13 +1,8 @@
 export { createConfig, LoggerConfig } from "./config-type";
-export {
-  AbstractEndpoint,
-  EndpointInput,
-  EndpointOutput,
-  EndpointResponse,
-} from "./endpoint";
```
RobinTail, line 7: ⚠️ breaking

src/index.ts
```diff
@@ -1,13 +1,8 @@
...
+export { AbstractEndpoint } from "./endpoint";
 export { Method } from "./method";
 export { EndpointsFactory, defaultEndpointsFactory } from "./endpoints-factory";
-export { IOSchema, FlatObject, markOutput } from "./common-helpers";
```
RobinTail, line 10: ⚠️ breaking

RobinTail (commented):
Ready

PR #423, opened 2022-04-30
Deprecated: Utility types

RobinTail:
Marking the entities that are going to be removed later in #408 as deprecated.


RobinTail (commented):
Ready

PR #483, opened 2022-06-24
Feat: a hook

RobinTail:
I'd like to provide you with a comprehensive React hook solution.

Key features:
- supports React 16.x and 17.x
  - I'm going to support 18.x, but hook testing library still does not work well with it
- integration with a generated client
- conditional requests
- watching for variable updates
- refresh on demand
- cache
  - with a TTL
  - customizable implementation
- with a customizable error handling


RobinTail:
this actually has to be either another package or a part of the generated code, along with the client.

PR #501, opened 2022-07-16
Fix: calling logger methods without a message

RobinTail:
There is an issue with logger when calling its methods without a message:
```typescript
logger.debug({something: true});
```
The output was empty, considering the first argument to be a message (string).
This PR is fixing it by adding `[No message]` message string before printing the object.

I'm also improving the way of testing the logger.


RobinTail:
ESM test fails on Node 18, only in CI

RobinTail:
Node 18 version in CI is 18.6.0

RobinTail:
OK. There is a `ts-node` compatibility issue with Node 18.6.
Need to fix it in another PR first

RobinTail:
related issue: https://github.com/TypeStrong/ts-node/issues/1839

RobinTail:
going to fix it in the following PR:
https://github.com/RobinTail/express-zod-api/pull/502

Since the issue is not fixed yet in the latest version, I'm going to restrict Node 18.x to the previous version 18.5

RobinTail (commented):
ready

PR #571, opened 2022-08-28
Feat: Tagging endpoints

RobinTail:
See #564 

image


pupca:
@RobinTail any progress on this experiment? What is missing? Any help needed?

RobinTail:
@pupca , what's missing is an approach for enforcing constraints on tags used across the API, especially when using different EndpointFactories.
I believe this is required in order to avoid the manual input mistakes when tagging the endpoints.
Perhaps there should be a method of the Factory for setting all possible tags from an array, but it have to be applied to all Factories used in the API.
In case you have better ideas, please let me know or create a PR.

RobinTail:
@pupca , check out the updated PR and let me know what you think.
Thank you

RobinTail:
It would be great to have a list of the tags along with their descriptions in the configuration.
However, currently, EndpointsFactory does not interact with Config.
Moreover, the library provides `defaultEndpointsFactory`, a ready-to-go instance that does not need any configuration.
The workaround could be some method that applies the configuration to the factory... but I'm sure I like this.
@pupca

RobinTail:
@pupca , you know what... I have an idea :)
Tagging will be a feature on demand. So, we'll keep the `defaultEndpointsFactory` as it is, without tags.
However, we'll make the `EndpointsFactory` constructor to accept the configuration (maybe even optionally), along with the ResultHandler.
This way, we could keep the tags in a one single place — the configuration, but you'll have to opt-in for using tags by creating your own instance of `EndpointsFactory`.

pupca:
Hi @RobinTail, first of all thanks for taking your time implement this feature! 

Hm, I see what you are doing there and yes that can be an option. But perhaps it could be done in very simplistic way, that in the `OpenAPI` class you will define the tags and its descriptions and in the `EndpointsFactory` you will just be using the string representations (ids) of that tags. Yes, sure, refactoring of tags or missing/extra tags will be 'hard' to spot, but I think that is quite acceptable.

As well as I think we should add the `summary` tag into the `defaultEndpointsFactory` ([see usage here](https://swagger.io/docs/specification/paths-and-operations/)) as this is quite a important part of the endpoint description in OpenApi, what do you think?

RobinTail:
> As well as I think we should add the summary tag into the defaultEndpointsFactory ([see usage here](https://swagger.io/docs/specification/paths-and-operations/)) as this is quite a important part of the endpoint description in OpenApi, what do you think?

Yes, this is exactly what I'd like to achieve by keeping tags in the configuration object.
At the same I'd like to prevent manual tagging mistakes when creating endpoints, by enforcing constraints if tags are present in the configuration.
I think I'm getting closer to the solution that would fulfill all these needs by using `EndpointsFactory` constructor overloads.
That would preserve the backward compatibility as well.
Thank you for your support and comments @pupca

RobinTail:
Nice!

image

pupca:
Looks good @RobinTail , would you be able to squeeze in the the `summary` tag, perhaps in the same fashion as the `description` tag ?

src/endpoints-factory.ts
```diff
@@ -29,37 +30,57 @@ type BuildProps<
...
+    config?: CommonConfig<TAG>;
+  });
+  constructor(
+    subject:
+      | ResultHandlerDefinition<POS, NEG>
+      | {
+          resultHandler: ResultHandlerDefinition<POS, NEG>;
+          config?: CommonConfig<TAG>;
+        }
+  ) {
```
RobinTail, line 67: Oh, yes! Now I like it :)

RobinTail:
> Looks good @RobinTail , would you be able to squeeze in the the `summary` tag, perhaps in the same fashion as the `description` tag ?

Oh, I see what you mean.
You'd like to have `summary` prop along with the `description` one when creating endpoint.
Yeah, sure, it's possible, but I'm not sure it's related to the tagging feature, but I'm going think on it for sure. @pupca

pupca:
Right, that (tags and summary) just make the OpenApi feature way more usable, therefore i kinda put it together :D

RobinTail:
Regarding this `summary` prop, my concern is that I'd like to avoid an uncertainty on which property to use: `description` or `summary`. Maybe it's obvious for advanced OpenAPI users, but definitely not for beginners. Perhaps, I'm going to make a prop with a name like `shortDescription` that would be assigned to the `summary` in the OpenAPI spec.
What do you think, @pupca ?

pupca:
@RobinTail, yes that certainly work, or make the prop as like OpenApi object, so instead of 'description' (which can stay for backwards compatibility), we can have `specification` object that would have props `{ description: string; summary: string }` and perhaps many more from the OpenApi definition if anyone would want.

pupca:
> Upd.
And maybe I'd copy description to the summary if it's short enough. But how exactly short — I don't know yet.

Just watch out, those are two different things in the OpenApi definition, `summary` which is one sentence description of the endpoint like `Get collection of users` and `description` which might be a longer text talking about the endpoint more in depth.

src/endpoints-factory.ts
```diff
@@ -29,37 +30,57 @@ type BuildProps<
...
-  SCO extends string = string
+  SCO extends string = string,
+  TAG extends string = string
 > {
+  protected resultHandler: ResultHandlerDefinition<POS, NEG>;
   protected middlewares: AnyMiddlewareDef[] = [];
 
-  constructor(protected resultHandler: ResultHandlerDefinition<POS, NEG>) {}
+  constructor(resultHandler: ResultHandlerDefinition<POS, NEG>);
+  constructor(params: {
```
RobinTail, line 56: add jsdoc here

RobinTail (commented):
Ready. Need to describe the feature in Readme and Changelog.

RobinTail (commented):
Ready

pupca:
👍

RobinTail:
Released in v8.1.0 🚀

RobinTail:
@pupca , the `summary` feature released in v8.2.0 🚀

PR #582, opened 2022-09-04
Breaking: Removing the deprecated signature 

RobinTail:
the deprecated case of EndpointsFactory::addMiddleware() with using MiddlewareCreationProps as an argument

### DIFF
--- a/src/endpoints-factory.ts
+++ b/src/endpoints-factory.ts
@@ -15,7 +15,6 @@ import {
   createMiddleware,
   ExpressMiddleware,
   ExpressMiddlewareFeatures,
-  MiddlewareCreationProps,
   MiddlewareDefinition,
 } from "./middleware";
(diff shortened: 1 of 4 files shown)


src/endpoints-factory.ts
```diff
@@ -71,52 +70,14 @@ export class EndpointsFactory<
...
-    SCO & ASCO
-  >;
-
-  /** @deprecated please use createMiddleware() for the argument */
-  public addMiddleware<
-    AIN extends IOSchema<"strip">,
-    AOUT extends FlatObject,
-    ASCO extends string
-  >(
-    props: MiddlewareCreationProps<AIN, OUT, AOUT, ASCO>
```
RobinTail, line 90: ⚠️ Breaking

RobinTail (commented):
ready

PR #589, opened 2022-09-06
Handling non-Error exceptions

### LINKED ISSUE #585: improve handling of non-object errors thrown in endpoint handlers
oh boy, i just spent the last couple hours running in circles trying to figure this one out. my API endpoint was returning `{"status":"success"}` with no data, which was then causing errors when my generated client was assuming `"status":"success"` means `data` should be present

the culprit is this:
1. the error is only passed to the result handler if it's `instanceof Error` - https://github.com/RobinTail/express-zod-api/blob/721a41e2f762a9f0a524fc111549c461d9057462/src/endpoint.ts#L333-L335
2. if there's no error passed to the result handler, it returns a 200 and a success status regardless of whether there's any real data or not - https://github.com/RobinTail/express-zod-api/blob/721a41e2f762a9f0a524fc111549c461d9057462/src/result-handler.ts#L83-L89
3. turns out that running `prettier` programmatically throws a `string` 😅 - this was in my code

I was about to open a PR up but I'm actually not really sure how you want to handle this since this will be a breaking change, albeit a needed one IMO (probably before the v8 release?). the two obvious choices seem to be:
1. widening the typings on the `ResultHandler` so you can pass in any kind of error (thus passing the onus to the default handler or the downstream consumer to handle more error cases)
2. just stringify any non-object error and create an Error object out of it

the latter seems way easier and with less breaking changes, whereas the former will certainly break any custom result handler implementations. but if you're gonna make a breaking change anyway, I think the former is the way to go because if you just coerce everything into a string then it gives downstream consumers no way to implement custom error handling based on the type of the error

so my _suggestion_ is to change the error type in the `ResultHandler` to `unknown`, then check if it's `!== undefined` (or `!= null` to handle both undefined & null?) rather than just checking if it's falsey like you currently do. but not sure if you've got some other thoughts on how to better handle this?

RobinTail:
Closes #585

### DIFF
--- a/src/endpoint.ts
+++ b/src/endpoint.ts
@@ -184,7 +184,7 @@ export class Endpoint<
           })),
         ]);
       }
-      throw e;
+      throw e instanceof Error ? e : new Error(`${e}`);
     }
   }
 
@@ -272,13 +272,14 @@ export class Endpoint<
         input: initialInput,
       });
     } catch (e) {
-      if (e instanceof Error) {
-        lastResortHandler({
-          logger,
-          response,
-          error: new ResultHandlerError(e.message, error),
-        });
-      }
+      lastResortHandler({
+        logger,
+        response,
+        error: new ResultHandlerError(
+          (e instanceof Error ? e : new Error(`${e}`)).message,
+          error
+        ),
+      });
     }
   }
 
@@ -330,9 +331,7 @@ export class Endpoint<
         await this.#parseAndRunHandler({ input, options, logger })
       );
     } catch (e) {
-      if (e instanceof Error) {
-        error = e;
-      }
+      error = e instanceof Error ? e : new Error(`${e}`);
     }
     await this.#handleResult({
       initialInput,


src/endpoint.ts
```diff
@@ -184,7 +184,7 @@ export class Endpoint<
           })),
         ]);
       }
-      throw e;
+      throw e instanceof Error ? e : new Error(`${e}`);
```
rayzr522, line 187: i literally just wrote this exact line of code in my own repo to coerce prettier's fucky string errors LMAO

src/endpoint.ts
```diff
@@ -272,13 +272,14 @@ export class Endpoint<
...
-          logger,
-          response,
-          error: new ResultHandlerError(e.message, error),
-        });
-      }
+      lastResortHandler({
+        logger,
+        response,
+        error: new ResultHandlerError(
+          (e instanceof Error ? e : new Error(`${e}`)).message,
```
rayzr522, line 279:
```suggestion
          e instanceof Error ? e.message : `${e}`,
```

src/server.ts
```diff
@@ -46,13 +46,14 @@ export const createNotFoundHandler =
...
-          response,
-          logger,
-          error: new ResultHandlerError(e.message, error),
-        });
-      }
+      lastResortHandler({
+        response,
+        logger,
+        error: new ResultHandlerError(
+          (e instanceof Error ? e : new Error(`${e}`)).message,
```
rayzr522, line 53:
```suggestion
          e instanceof Error ? e.message : `${e}`,
```

rayzr522 (changes requested):

src/endpoint.ts
RobinTail: Yeah, that definitely can be simplified, but I also had an idea of extracting the `unknown` exception handling into a function like this:

```ts
function makeErrorFromAnything(subject: unknown) {
  return e instanceof Error ? e.message : `${e}`;
}
```

so that stringifying behaviour would be kept in a single dedicated place and could be adjusted or even customized in the future.

tests/unit/common-helpers.spec.ts
```diff
@@ -629,4 +632,43 @@ describe("Common Helpers", () => {
...
+        `[\n  {\n    "code": "invalid_type",\n    "expected": "string",\n` +
+          `    "received": "number",\n    "path": [\n      ""\n` +
+          `    ],\n    "message": "invalid type"\n  }\n]`,
+      ],
+      [undefined, "undefined"],
+      [null, "null"],
+      ["string", "string"],
+      [123, "123"],
+      [{}, "[object Object]"],
+      [{ test: "object" }, "[object Object]"],
```
RobinTail, line 658: Todo: I don't like it

RobinTail:
"Stringify anything"
https://gist.github.com/KaptajnKold/1000978

tests/unit/common-helpers.spec.ts
RobinTail: Actually, I think it's fine.
Just don't pass the bullshit to the `throw` statement :)

src/endpoint.ts
RobinTail: it turned out that `Symbol` require special handling

RobinTail:
@rayzr522 , please review

RobinTail:
Ok. Since it's taking a bit longer than I expected, I'm going to merge it anyway.
I hope it's OK for you @rayzr522 .
In case something isn't right, feel free to open another issue.

rayzr522:
> Ok. Since it's taking a bit longer than I expected, I'm going to merge it anyway. I hope it's OK for you @rayzr522 . In case something isn't right, feel free to open another issue.

sorry for the late reply, didn't see this!! looks fantastic, thank you so so much for taking care of this so quickly :3

PR #643, opened 2022-10-13
OpenAPI generator: Deny using z.file() in input.

RobinTail:
`z.file()` was introduced in v2.1.0.
It was not intended to be used in endpoint I/O.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 8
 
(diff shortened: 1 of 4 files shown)


RobinTail:
Since the interface is not changed, while the behaviour changed, I'm going to consider this as a `patch` according to SemVer.

RobinTail (commented):
Should be fine for now.

PR #652, opened 2022-10-19
Silence zod parse error

hunght:


tests/unit/endpoint.spec.ts
```diff
@@ -513,6 +513,40 @@ describe("Endpoint", () => {
...
     });
 
+    test("silence zod parse error option", async () => {
+      const endpoint = defaultEndpointsFactory.build({
+        method: "post",
+        input: z.object({
+          limit: z.preprocess(
+            (a) => parseInt(z.string().parse(a), 10),
+            z.number().positive().max(100)
+          ),
```
RobinTail, line 523: Instead all that, you just need more loose schema in your project, like this one:

```ts
z.number().or(
  z.string().transform(
    (str) => parseInt(str, 10)
  )
)
```

RobinTail: this way, you accept both `number` and `string`, but in any case you supply `number` to your handler.

RobinTail:
What this PR does is disabling the validation, which is one of the primary goals of the library.
I believe that this does not add value for the users of the library, but it goes against the conceptual intent.
I think the approach should be different, but if still that's what you want for your needs, you can make a fork with this option — the library is open source under MIT license.

PR #655, opened 2022-10-20
Feat: Top level refinements

### LINKED ISSUE #654: Refinements on input object
Currently, there is no way to do refinements on the entire input object, only on it's properties. It would be great if there was a way to support this:

example:

body is
```
{
  type: 'type1',
  dynamicValue: { 
     type1Attribute: 1,
  } 
}
```
```
{
  type: 'type2',
  dynamicValue: { 
     type2Attribute: '1',
  } 
}
```

I want to be able to do:

```
   input: z.object({ 
      type: z.union([z.literal('type1'), z.literal('type2')]),
      dynamicValue: z.union([
        z.object({ type1Attribute: z.number() }),
        z.object({ type2Attribute: z.string() })
      ]),
    }).refine(data => {
       if (data.type === "type1") {
          return !!(data.dynamicValue as any).type1Attribute
       }
       return true;
    }, { message: 'type1Attribute is required if type is type1', path: ['dynamicValue']]),
   ...
});
```

but I get an error like this..

`type ZodEffects<....> is not assignable to type IOSchema<any>`

As a workaround, I could change my input to be something like..
```
z.object({ data: z.object(...).refine(...) })
```

But I don't want to have to define my inputs like that.

RobinTail:
Closes #654 
Fulfills #600


RobinTail:
I found an issue in Zod, that I'd like to be fixed before I continue with this feature.
Made a PR to fix it.
https://github.com/colinhacks/zod/pull/1513

tests/unit/endpoint.spec.ts
```diff
@@ -544,4 +544,122 @@ describe("Endpoint", () => {
...
+      const { responseMock } = await testEndpoint({
+        endpoint,
+        requestProps: {
+          method: "POST",
+          body: {},
+        },
+      });
+      expect(responseMock.json).toHaveBeenCalledWith({
+        error: {
+          message: ": Please provide at least one property", // @todo can we fix the missing field?
```
RobinTail, line 658: The issue was discovered while trying to fix this

RobinTail (commented):
Ready 🤞🏽 
Need to describe the feature.

RobinTail (commented):
Ready

PR #656, opened 2022-10-21
Fix warning on deprecated `set-output` in CI

RobinTail:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Partially fixed, except Coveralls actions. Related issue: https://github.com/coverallsapp/github-action/issues/135


RobinTail (commented):
ready

PR #674, opened 2022-10-28
Bug: transformations in middlewares

### LINKED ISSUE #673: Bug: Cannot use `dateIn` as an input for a middleware
Another small bug my team discovered:

Right now, it's not possible to define a middleware with a `dateIn` input. Calling an endpoint using such a middleware always results in a `<field_name>: Expected string, received date` error.

I didn't have time to look into the source code here, but it seems like inputs might be parsed/validated multiple times (once at the middleware and then again at the endpoint), and by the time they're checked for the second time the date string has already been parsed to a date (which is not the expected input type of `dateIn`). Could that be the case?

RobinTail:
Fixes #673

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Version 8
 
+### v8.3.2
+
+- Fixed the bug #673 found and reported by [@shroudedcode](https://github.com/shroudedcode).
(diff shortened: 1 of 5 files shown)


RobinTail (commented):
Ready

RobinTail (commented):
Ready

PR #676, opened 2022-10-28
Fix: setting description for `dateIn` and `dateOut`

### LINKED ISSUE #672: Bug: Cannot set description on `dateIn` and `dateOut` fields
First of all: Great job on this library! It manages to strike an almost perfect balance between being super quick and easy to get started with while also not being too restrictive or opinionated. Kind of like tRPC but more "public-API-friendly".

Anyway, my team and I discovered a small bug in how the OpenAPI schema is being generated:

All fields with the `dateIn` or `dateOut` type don't keep the descriptions assigned using `.describe()`. Instead, they always have the description set to `YYYY-MM-DDTHH:mm:ss.sssZ`. It seems to be caused by [the `depictDateIn` and `depictDateOut` functions](https://github.com/RobinTail/express-zod-api/blob/1ab41c42b6a8ff7a95bc611015e8095130241faf/src/open-api-helpers.ts#L209-L244) not taking into account custom descriptions.

RobinTail:
Fixes #672

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Version 8
 
+### v8.3.3
+
+- Fixed the bug #672 found and reported by [@shroudedcode](https://github.com/shroudedcode).
+  - Preserving the custom description of `z.dateIn()` and `z.dateOut()` schemas when generating OpenAPI documentation.
+
+```yaml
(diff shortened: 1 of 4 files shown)


RobinTail (commented):
Ready

PR #678, opened 2022-10-30
Adjusting top level refinements

RobinTail:
In some cases the type of refinement can be indistinguishable from the type of transformation.
Since both of them are using the same class `ZodEffects`, the only difference is the inequality if input and output types.
However, both of these types may have a common ancestor, which make it challenging to recognize them on the level of Types.

Example:

```ts
// ZodEffects<ZodObject<{}>, boolean, {}>
z.object({}).transform(() => true);  // OK, this is catchable
// ZodEffects<ZodObject<{}>, never[], {}>
z.object({}).transform(() => []);  // never[] inherits Array inherits Object, {} inherits Object as well
```

So, at this point I made a decision to handle this case programmatically.

In case anyone smarter than me is reading this, please let me know how I can improve `IOSchema` type to allow refinements and exclude transformations. ¯\_(ツ)_/¯


RobinTail (commented):
Ready for changelog

PR #696, opened 2022-11-18
Testing the case of the discussion 694.

RobinTail:
Related to the discussion #694

### DIFF
--- a/tests/unit/__snapshots__/open-api.spec.ts.snap
+++ b/tests/unit/__snapshots__/open-api.spec.ts.snap
@@ -2074,6 +2074,9 @@ paths:
                     enum:
                       - success
                   data:
+                    example:
+                      "null": null
+                      dateOut: 2022-08-04
                     type: object
                     properties:
                       "null":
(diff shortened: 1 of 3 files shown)


tests/unit/__snapshots__/open-api.spec.ts.snap
```diff
@@ -2074,14 +2074,17 @@ paths:
                     enum:
                       - success
                   data:
+                    example:
+                      "null": null
+                      dateOut: 2022-08-04
```
RobinTail, line 2079: the transformation is applied here

tests/unit/__snapshots__/open-api.spec.ts.snap
```diff
@@ -2092,6 +2095,13 @@ paths:
                 required:
                   - status
                   - data
+              examples:
+                example1:
+                  value:
+                    status: success
+                    data:
+                      "null": null
+                      dateOut: 2022-08-04
```
RobinTail, line 2104: and here

RobinTail:
The case is closed — there is actually no issue.

PR #702, opened 2022-11-19
Making Typescript a regular dependency

RobinTail:
Operating TS factories should depend on particular version of Typescript.
Either it has to be defined as a peer in a more strict way, OR Typescript has just to be a regular dependency (not dev+peer).

Similar PR to `zod-to-ts` https://github.com/sachinraja/zod-to-ts/pull/24 , issue https://github.com/sachinraja/zod-to-ts/issues/23


RobinTail (commented):
ready

PR #711, opened 2022-11-25
Initial evaluation of schema refs

VideoSystemsTech:
OpenAPI client generators often create classes/types for requests and responses objects by looking at the `schemas` field in the `components` object ([reference](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#components-object)).

I drafted this PR that adds a `schema` field to the `withMeta()` method: with this meta, the schema is added to the global OpenAPI document (under `components`), and in the path the schema is [referenced](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#reference-object).

This doesn't consider:  
- nested schemas
- schema name uniqueness

### DIFF
--- a/example/endpoints/update-user.ts
+++ b/example/endpoints/update-user.ts
@@ -19,20 +19,24 @@ export const updateUserEndpoint =
         name: z.string().min(1),
         birthday: z.dateIn(),
       })
-    ).example({
-      id: "12",
-      name: "John Doe",
-      birthday: "1963-04-21",
-    }),
+    )
+      .example({
+        id: "12",
+        name: "John Doe",
+        birthday: "1963-04-21",
+      })
+      .schema("UpdateUserRequest"),
     output: withMeta(
       z.object({
         name: z.string(),
         createdAt: z.dateOut(),
       })
-    ).example({
-      name: "John Doe",
-      createdAt: new Date("2021-12-31"),
-    }),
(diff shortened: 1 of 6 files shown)


RobinTail:
@VideoSystemsTech . Ok, thank you. But why it might be needed?

RobinTail:
@VideoSystemsTech , still waiting for your answer.
What is the benefit of having references instead of having schemas in place?

VideoSystemsTech:
@RobinTail Hi, [OpenAPI Generator](https://openapi-generator.tech/) is used to generate client libraries from an `openapi.json` file.  
When you generate a client library, most of the time it will create a class/interface/type in your desired programming language with the name provided in the $ref.
For example, [this pet store specification](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/test/resources/3_0/petstore.yaml) has a bunch of references (Category, User, Tag, etc.), and the generator can generate classes from those names. [Here](https://github.com/OpenAPITools/openapi-generator/tree/master/samples/client/petstore/typescript-fetch/builds/default/models) are the generated TypeScript models for the pet store specification.

RobinTail:
After doing #856 it becomes easier to implement this feature and make it optional.
I'm going to research this opportunity soon.

PR #728, opened 2022-12-09
Fix overly strict ISO date validation

shroudedcode:
A customer had trouble calling one of our API endpoints because their request client (Java Spring `RestTemplate`) generates ISO dates where the sub-second digits aren't (always?) three. Instead of `2022-12-09T08:50:35.830Z`, their library would submit `2022-12-09T08:50:35.83Z`. This caused our API to return a 400 error because the validation of the relevant `dateIn` would fail.

I updated the relevant regex to allow any number of digits which seems to be valid [according to this](https://www.w3.org/TR/NOTE-datetime#:~:text=one%20or%20more%20digits%20representing%20a%20decimal%20fraction%20of%20a%20second) and is parsed perfectly fine by `new Date()`.

### DIFF
--- a/src/date-in-schema.ts
+++ b/src/date-in-schema.ts
@@ -12,6 +12,7 @@ import { isValidDate } from "./common-helpers";
 
 // simple regex for ISO date, supports the following formats:
 // 2021-01-01T00:00:00.000Z
+// 2021-01-01T00:00:00.0Z
 // 2021-01-01T00:00:00Z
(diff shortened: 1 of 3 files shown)


RobinTail:
Nice

RobinTail (approved):

RobinTail (approved):

RobinTail:
🚀  released in v8.4.3, @shroudedcode

PR #734, opened 2022-12-09
Switching from Coveralls to Codecov

RobinTail:


RobinTail:
no more need for this, due to https://github.com/RobinTail/express-zod-api/pull/844

PR #758, opened 2022-12-23
Wrong listing of required object props in case of coercion in response

RobinTail:
Found a bug introduced in 8.6.0


RobinTail (commented):
Ready

PR #794, opened 2023-01-30
Update input validation to explicitly throw an input validation error on failure

TheWisestOne:
- Add a common helper function `parseInput` that will wrap an attempt to parse a given input in a `try/catch` and throw a 400 http error in the event of failure
 - Use this new helper function when parsing inputs for middleware and handlers
 - Update `getStatusCodeFromError` to remove now-unnecessary `ZodError` check
 - Update `getStatusCodeFromError` to correctly check if an error is an HttpError (according to the readme of `http-error`: https://github.com/jshttp/http-errors )
 
 this relates to this discussion: https://github.com/RobinTail/express-zod-api/discussions/787


RobinTail:
Ok. So, basically, it makes `ZodError` to be transformed into HttpError with a status code `400` when it's about the input validation, right?

I see some test were removed. Specifically `ResultHandler` ones.
I'd like them to be rather adjusted. And probably the `defaultResultHandler` needs some changes as well.

By the way. If your database I/O validation (that utilizes zod for this purpose) will throw `ZodError`, how these changes would help it to respond with `500`? I'm not sure I did understand the handling of this particular scenario. 

But in general, I think it's a good move in the right direction, @TheWisestOne . Thank you for that.

RobinTail:
There is one more idea came to my mind later.
By doing this transformation (ZodError to HttpError) in `Endpoint`, we actually take away the ability of the customization of the ZodError formatting from the developer. I mean the conversion of `ZodIssue` array inside `ZodError` into a string `message` of HttpError.
I would prefer to avoid this, letting the developer to be able to customize it according to their needs.
Perhaps we should think a bit on possible extraction of some entity into configuration.

TheWisestOne:
Yup, exactly!

As for `ResultHandler` tests, I explicitly removed a test case because I indirectly removed some logic from the `defaultResultHandler` (the `if` statement related to zod errors in `getStatusCodeForError`). Since it no longer has this logic, it no longer needs that test case. The logic for getting the correct status code for an http error is already covered by the `Should handle HTTP error` test case).

`defaultResultHandler` does not need any explicit changes because my changes to `getStatusCodeFromError` will affect that as well, so it "just works" now.

The removal of the `if` statement in `getStatusCodeForError` is the one that allows internal Zod errors to still be returned as 500s. In the old structure, any internally thrown `ZodError` would be transformed into a 400 by `getStatusCodeForError`. By removing that logic, we allow any internally-thrown `ZodError` to correctly become a 500.

TheWisestOne:
> There is one more idea came to my mind later. By doing this transformation (ZodError to HttpError) in `Endpoint`, we actually take away the ability of the customization of the ZodError formatting from the developer. I mean the conversion of `ZodIssue` array inside `ZodError` into a string `message` of HttpError. I would prefer to avoid this, letting the developer to be able to customize it according to their needs. Perhaps we should think a bit on possible extraction of some entity into configuration.

This is a good point, and at first I thought I perhaps broke your `getMessageFromError` logic as a result.
However, when I looked into this further, it turns out that calling `createError(statusCode, error)` does not actually change the `ZodError`, but rather extends it: https://github.com/jshttp/http-errors#createerrorstatus-error-properties

As a result, your existing logic in `getMessageFromError` actually still works. This can be validated by the fact that the unit test in `endpoint.spec.ts` titled: `"should fail during the refinement of invalid inputs"` passes without any changes, so the error customization *does* in fact work with the current implementaiton.

TheWisestOne:
> This is a good point, and at first I thought I perhaps broke your `getMessageFromError` logic as a result. However, when I looked into this further, it turns out that calling `createError(statusCode, error)` does not actually change the `ZodError`, but rather extends it: https://github.com/jshttp/http-errors#createerrorstatus-error-properties
> 
> As a result, your existing logic in `getMessageFromError` actually still works. This can be validated by the fact that the unit test in `endpoint.spec.ts` titled: `"should fail during the refinement of invalid inputs"` passes without any changes, so the error customization _does_ in fact work with the current implementaiton.

I have just updated this PR with unit tests for `parseInput` which showcase the behaviour I'm discussing here

TheWisestOne:
@RobinTail please let me know if you have any more questions / concerns about this PR

TheWisestOne:
@RobinTail just bumping this in case it slipped down from your queue. Let me know if there's anything about this PR that you still have concerns about. I believe in its current state it is ready for a final review

RobinTail:
@TheWisestOne , sorry for delay.
It's a bit much things happening.
I keep in mind this request and I'm going to get back to it soon.
We will find a good solution.

TheWisestOne:
> @TheWisestOne , sorry for delay. It's a bit much things happening. I keep in mind this request and I'm going to get back to it soon. We will find a good solution.

 No worries, thanks for taking a look!

RobinTail:
Thank you very much @TheWisestOne for all your efforts.
In end I was inspired by your solution, so I rewrote my #819 .
I'm closing this one and the changes will be released in v9.0.0-beta2

Issue #805, opened 2023-02-09
API Routes with multiple `-` in the path breaks client library generation


TheWisestOne:
When an API route has multiple `-` in the path, for example `/api/test-this-one-out`, generating the client-side library breaks.

Specifically, the type names that are generated include all the `-` characters after the first one. My guess is there is a `replace` that should be a `replaceAll` or something along those lines

An example screenshot is attached
![Screen Shot 2023-02-09 at 4 20 36 PM](https://user-images.githubusercontent.com/83668621/217952489-752da87a-fa84-4d7c-b802-870056ba8943.png)

RobinTail:
Thank you @TheWisestOne .
The fix released in v8.9.1 🚀

PR #806, opened 2023-02-09
Fix client-helpers `cleanId` function to properly clean the entire string

TheWisestOne:
The current function will only replace the first occurence of an invalid character. This fixes it to handle multiple invalid characters. Also, update the unit test for `client.ts` to exercise this code.

This resolves https://github.com/RobinTail/express-zod-api/issues/805

### DIFF
--- a/src/client-helpers.ts
+++ b/src/client-helpers.ts
@@ -200,7 +200,8 @@ export const cleanId = (path: string, method: string, suffix: string) => {
   return [method]
     .concat(path.split("/"))
     .concat(suffix)
-    .map((entry) => entry.replace(/[^A-Z0-9]/gi, ""))
+    .map((entry) => entry.split(/[^A-Z0-9]/gi))
+    .flat(1)
     .map(
       (entry) => entry.slice(0, 1).toUpperCase() + entry.slice(1).toLowerCase()
     )
--- a/src/client-helpers.ts
+++ b/src/client-helpers.ts
@@ -200,7 +200,7 @@ export const cleanId = (path: string, method: string, suffix: string) => {
   return [method]
     .concat(path.split("/"))
     .concat(suffix)
-    .map((entry) => entry.replace(/[^A-Z0-9]/i, ""))
+    .map((entry) => entry.replace(/[^A-Z0-9]/gi, ""))
     .map(
       (entry) => entry.slice(0, 1).toUpperCase() + entry.slice(1).toLowerCase()
     )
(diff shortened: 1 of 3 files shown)


src/client-helpers.ts
```diff
@@ -200,7 +200,8 @@ export const cleanId = (path: string, method: string, suffix: string) => {
   return [method]
     .concat(path.split("/"))
     .concat(suffix)
-    .map((entry) => entry.replace(/[^A-Z0-9]/i, ""))
+    .map((entry) => entry.split(/[^A-Z0-9]/gi))
```
TheWisestOne, line 203: the addition of the `g` allows the clean function to affect all invalid characters
The use of `.split` followed by `.flat` allows for the `CapitalCasing` that happens next to correctly map these `special-character_separated` words into capital case correctly: `SpecialCharacterSeparated`

src/client-helpers.ts
```diff
@@ -200,7 +200,8 @@ export const cleanId = (path: string, method: string, suffix: string) => {
...
+    .flat(1)
```
RobinTail, line 204:
```suggestion
    .flat()
```

`1` [is the default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#parameters)

src/client-helpers.ts
RobinTail: ok

RobinTail: there is also `flatMap`, which is described are more efficient

> It is identical to a map() followed by a flat() of depth 1 , but slightly more efficient than calling those two methods separately.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

src/client-helpers.ts
```diff
@@ -200,7 +200,8 @@ export const cleanId = (path: string, method: string, suffix: string) => {
   return [method]
     .concat(path.split("/"))
     .concat(suffix)
-    .map((entry) => entry.replace(/[^A-Z0-9]/i, ""))
+    .map((entry) => entry.split(/[^A-Z0-9]/gi))
+    .flat()
```
RobinTail, line 204:
```suggestion
    .flatMap((entry) => entry.split(/[^A-Z0-9]/gi))
```

there is also `flatMap`, which is described as more efficient

> It is identical to a map() followed by a flat() of depth 1 , but slightly more efficient than calling those two methods separately.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

RobinTail (approved):
Great!

PR #819, opened 2023-02-14
The proposed solution to problem #787

RobinTail:
Related to the discussion #787 

Should enable the developer using `zod` as a tool within Endpoint's handler without considering its possible parsing/validation errors as IO related ones, leading to the status code `500` instead of `400` in that case.

@TheWisestOne 

### The approach

1. I am making an `InputValidationError`, consistent with `OutputValidationError`.
2. I separate the input parsing from the execution.
3. Validation errors (ZodError) catched before running middlewares and Endpoint handlers are transformed into `InputValidationError`
  - It also contains `originalError` prop for the further processing if needed.
  - However the `message` prop is already prepared for being human readable.
5. Thus, your `ResultHandler` now can be sure that any ZodError is actually NOT input validation error and can respond with `500`.

Inspired by #794.

### ToDo

- needs more tests
- need to expose error processing helpers in index ✅


RobinTail:
@TheWisestOne , please review

src/endpoint.ts
```diff
@@ -285,12 +286,25 @@ export class Endpoint<
...
+        input: finalInput,
+        options,
+        logger,
+      });
+    } catch (e) {
+      if (e instanceof z.ZodError) {
+        // Discussion #787
+        // if Zod used in the Endpoint's handler implementation as tool for checking something not I/O-related
+        // we're transforming the possible parsing error into a custom one for the further handling by ResultHandler
+        throw new EndpointHandlerError(getMessageFromError(e));
```
TheWisestOne, line 304: After this point, any handler that processes `EndpointHandlerError` only has access to whatever is returned from `getMessageFromError`. From looking at the implementation there and how it handles `ZodError`, I think this is fine

TheWisestOne (commented):
Seems good to me!

I have 2 comments:
1. EndpointHandlerError is a very generic name, and might be better renamed `EndpointHandlerZodError` or something like that to clarify that any *other* errors thrown by the endpoint *don't* get wrapped in this error class

2. Now, the flow of execution looks like:

 -  parse middleware input / run middleware -> any errors go to ResultHandler directly
 - parse handler input -> any errors go to ResultHandler directly
 - run endpoint handler -> any zod errors get wrapped in `EndpointHandlerError`
 - parse output -> any zod errors get wrapped in `OutputValidationError`

It seems a little arbitrary when we wrap up errors into these wrapper classes vs. let the result handler deal with the raw error. I am fine with the current state, just want to raise that it is not all that consistent, and it might be worth thinking about if there's a way to make it more consistent.

RobinTail:
> EndpointHandlerError is a very generic name, and might be better renamed EndpointHandlerZodError or something like that to clarify that any other errors thrown by the endpoint don't get wrapped in this error class.

@TheWisestOne, yes, agree. That need to be fixed.

src/endpoint.ts
RobinTail: I think I can equip this error with a public property that contains the original ZodError

RobinTail: fixed in 67f7b49

RobinTail:
> and in the catch block we would throw a new `InputValidationError`

there was a reason why I try to refrain from making a validation error — it will be a breaking change for existing custom `ResultHandler`s made by developers.
Since `ZodError` is for a long time works as a synonym for the I/O validation error. 

But I also with your point on consistency.

Need to think on it a little bit.
Sorry that it takes a bit long.

TheWisestOne:
> > and in the catch block we would throw a new `InputValidationError`
> 
> there was a reason why I tried to refrain from making another validation error — it would be a breaking change for existing custom `ResultHandler`s made by developers. Since `ZodError` is for a long time works as a synonym for the input validation error.
> 
> But I also agree with your point on consistency.
> 
> Need to think on it a little bit. Sorry that it takes a bit long.

ah, an interesting point. One thing to note is that my attempt at this fix (https://github.com/RobinTail/express-zod-api/pull/794) actually is non-breaking for people using an `instanceof ZodError` check in their ResultHandler. 
This is because:
```     throw createHttpError(400, makeErrorFromAnything(e)); ``` still creates an error that is an instance of a `ZodError` if `e` was a zod error. I still see some of the benifits of your approach, but just wanted to point out that we might be able to do something non-breaking for existing `ResultHandler`s

**Comment by RobinTail** [comment]
I'm going to tackle this problem once again keeping in mind that the breaking change is acceptable, since I'm going to make v9 with another one.

**Comment by RobinTail** [comment]
I wanna go with the `InputValidationError`, for of consistency.

**Comment by RobinTail** [comment]
I wanna leave the utilization of `createHttpError` for the developer's decision on the final implementation and customizations.

File: src/endpoint.ts [hunk 2]
```diff
@@ -222,16 +222,10 @@ export class Endpoint<
     try {
       return await this.outputSchema.parseAsync(output);
     } catch (e) {
-      const error =
-        e instanceof z.ZodError
-          ? new z.ZodError(
-              e.issues.map(({ path, ...rest }) => ({
-                ...rest,
-                path: (["output"] as typeof path).concat(path),
```
RobinTail, line 230: this trick is moved to the helper

RobinTail:
OK. Now I'm happy with it.
I hope it will also fulfills your needs, @TheWisestOne 

Anyway, I tested that throwing ZodError inside the Endpoint is not considered as IO validation issue.
So you could use zod for your database validations.

PR #838, opened 2023-02-26
The `v9` is for Brianna Ghey

### LINKED ISSUE #820: `DELETE` endpoints with `requestBody` leads to an OpenAPI semantic error

RobinTail:
image

> On 11 February 2023, Brianna Ghey, a 16-year-old British transgender girl from Birchwood in Warrington, Cheshire, England, was killed in Culcheth Linear Park in Culcheth, Warrington. Two accused teenagers, a boy and a girl both aged 15, are in police custody, having been charged with murder four days after the incident. A motive has not been established and police are investigating "all lines of inquiry", including investigating the killing as a possible hate crime

https://en.wikipedia.org/wiki/Killing_of_Brianna_Ghey

Closes #820


RobinTail:
🚀  v9.0.0-beta1 released, tag `beta`

src/common-helpers.ts
```diff
@@ -98,14 +100,18 @@ export function getMessageFromError(error: Error): string {
       )
       .join("; ");
   }
+  if (error instanceof OutputValidationError) {
+    const hasFirstField = error.originalError.issues[0].path.length > 0;
```
RobinTail, line 104: `issues` actually may be empty.
need to handle this case

RobinTail: fixed

RobinTail:
🚀  v9.0.0-beta2 released, tag `beta`

RobinTail:
🚀  v9.0.0-beta3 released, tag `beta`

RobinTail:
Now I wanna do a cleanup and refactoring of `Endpoint` and remove `createApiResponse()`

src/client.ts
```diff
@@ -46,14 +46,14 @@ export class Client {
...
+          schema: endpoint.getSchema("input"),
           isResponse: false,
         });
         const response = zodToTs({
           isResponse: true,
           schema: endpoint
-            .getPositiveResponseSchema()
-            .or(endpoint.getNegativeResponseSchema()),
+            .getSchema("positive")
+            .or(endpoint.getSchema("negative")),
```
RobinTail, line 56: interesting...

RobinTail (commented):
Nice, so far.

RobinTail:
🚀 v9.0.0-beta4 released, tag beta

.github/workflows/codeql-analysis.yml
```diff
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, v7 ]
+    branches: [ master, v8 ]
```
RobinTail, line 16: v8 branch created

RobinTail (commented):
should be fine 🤞🏽

PR #839, opened 2023-02-26
Add operationId to Endpoint

lazylace37:
Expose operationId to the OperationObject


RobinTail:
@lazylace37 , 
this approach would not work well, because `operationId` has to be unique, but you're making them set on the `Endpoint` level, which can handle several `methods`.

Would you mind if I'll make the `operationId` automatically generated from the Routing path and method?

lazylace37:
Hi @RobinTail, you are correct and no, I wouldn't mind.
Thanks

RobinTail:
The alternative solution is in #845

RobinTail:
The feature released in version 8.10.0 and 9.0.0-beta3, @lazylace37

PR #841, opened 2023-02-27
Custom response status

McMerph:
Related to #824 

This PR adds the ability to use custom response status in both `ResultHandler` and `OpenAPI` classes.

### DIFF
--- a/src/result-handler.ts
+++ b/src/result-handler.ts
@@ -20,6 +20,7 @@ interface ResultHandlerParams<RES> {
   request: Request;
   response: Response<RES>;
   logger: Logger;
+  positiveStatusCode: number;
 }
 
 type ResultHandler<RES> = (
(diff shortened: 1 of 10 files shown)


src/result-handler.ts
```diff
@@ -76,9 +78,17 @@ export const defaultResultHandler = createResultHandler({
...
   },
-  handler: ({ error, input, output, request, response, logger }) => {
+  handler: ({
+    error,
+    input,
+    output,
+    request,
+    response,
+    logger,
+    positiveStatusCode,
```
RobinTail, line 88: I believe ResultHandler should not receive the code.
It should decide on its own how to respond and the code should have only declarative meaning, consumed by the documentation generator.

RobinTail:
I'm going to do it in a slightly different way in #846 .
However, I'm very grateful to you @McMerph for offering this idea and all the efforts you made to implement this solution.

src/result-handler.ts
sarahssharkey: @RobinTail I'm having the same issue and was also thinking about passing the success status code to the result handler, because how is the result handler supposed to know what the success status code is? I want to return a 201 when a resource has been created, but return a 200 when a resource is updated or fetched. How is the result handler supposed to know which status code to return? It would be nicer to define the success status code in each endpoint's implementation - how should I do this without passing the success code to the result handler?

PR #845, opened 2023-02-27
Adding `operationId` to the documentation

RobinTail:
Alternative solution to #839

### DIFF
--- a/src/open-api.ts
+++ b/src/open-api.ts
@@ -38,6 +38,17 @@ interface GeneratorParams {
 export class OpenAPI extends OpenApiBuilder {
   protected lastSecuritySchemaIds: Partial<Record<SecuritySchemeType, number>> =
     {};
+  protected lastOperationIdSuffixes: Record<string, number> = {};
+
+  protected ensureUniqOperationId(path: string, method: Method) {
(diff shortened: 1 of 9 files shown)


src/open-api.ts
```diff
@@ -84,6 +84,7 @@ export class OpenAPI extends OpenApiBuilder {
         inputSources,
       });
       const operation: OperationObject = {
+        operationId: makeCleanId(path, method),
```
RobinTail, line 87: perhaps we should ensure the uniqueness of this id

RobinTail: done

src/open-api.ts
```diff
@@ -84,6 +95,7 @@ export class OpenAPI extends OpenApiBuilder {
         inputSources,
       });
       const operation: OperationObject = {
+        operationId: this.ensureUniqOperationId(path, method),
```
RobinTail, line 98: should it be optional (disableable in config)?

RobinTail: Perhaps it's fine for now.
Should not affect the performance much.

RobinTail (commented):
Generally ready. Need to describe the feature.

PR #846, opened 2023-02-28
Customizable status codes

RobinTail:
Related to the discussion #824 
Alternative to #841


src/api-response.ts
```diff
@@ -3,15 +3,44 @@ import { MimeDefinition, mimeJson } from "./mime";
...
+  statusCode?: number;
 };
 
-export const createApiResponse = <S extends z.ZodTypeAny>(
+type ApiResponseCreationProps<S extends z.ZodTypeAny> = {
+  schema: S;
+  statusCode?: number;
+} & ({ mimeTypes?: [string, ...string[]] } | { mimeType?: string });
+
+export function createApiResponse<S extends z.ZodTypeAny>(
```
RobinTail, line 17: perhaps I can plan just getting rid of this function completely, solving all cases inside `Endpoint` methods

src/endpoint.ts
```diff
@@ -182,11 +185,34 @@ export class Endpoint<
...
   }
 
   public override getNegativeMimeTypes() {
-    return this.resultHandler.getNegativeResponse().mimeTypes;
+    const apiResponse = this.resultHandler.getNegativeResponse();
+    return "mimeType" in apiResponse && apiResponse.mimeType
+      ? [apiResponse.mimeType]
+      : "mimeTypes" in apiResponse && apiResponse.mimeTypes
+      ? apiResponse.mimeTypes
+      : [mimeJson];
```
RobinTail, line 204: DNRY

src/api-response.ts
```diff
@@ -1,17 +1,31 @@
...
-  mimeTypes: string[];
-};
+export type ApiResponse<S = z.ZodTypeAny> = {
+  schema: S;
+  /** @default 200 for a positive response, 400 for a negative response */
+  statusCode?: number;
+} & (
+  | {
+      /** @default [application/json] */
+      mimeTypes?: string[];
```
RobinTail, line 10: should be non-empty

RobinTail: fixed

src/api-response.ts
```diff
@@ -1,17 +1,34 @@
...
+   * */
+  statusCode?: number;
+} & (
+  | {
+      /** @default [ "application/json" ] */
+      mimeTypes?: [string, ...string[]];
+    }
+  | {
+      /** @default "application/json" */
+      mimeType?: string;
```
RobinTail, line 17: this does not actually work.
typescript fails to perform this check on the function result

src/endpoint.ts
RobinTail: fixed

src/endpoint.ts
```diff
@@ -169,24 +172,59 @@ export class Endpoint<
...
+    const { mimeTypes, mimeType } = subject;
+    return mimeType ? [mimeType] : mimeTypes || fallback;
+  }
+
   public override getPositiveMimeTypes() {
-    return this.resultHandler.getPositiveResponse(this.outputSchema).mimeTypes;
+    return this.#getMimeTypesFromApiResponse(
+      this.resultHandler.getPositiveResponse(this.outputSchema)
+    );
   }
```
RobinTail, line 206: perhaps builder can do this.
it already supplies input mime types.
so we can cache it.

RobinTail: I'm going to do the refactoring in v9

RobinTail (commented):
Generally ready

RobinTail (commented):
Tests! We need tests

RobinTail (commented):
I think it's complete now 🤞🏽

Issue #822, opened 2023-02-14
Locking package versions


jasonk:
This is basically the same issue as #178, just a new version of it.

Because express-zod-api specifies that it is only compatible with zod version 3.20.2 you will get errors if try to use it in an app that uses any other version of zod.

This is all that is needed to reproduce it:

```sh
mkdir repro && cd repro
yarn init -y
yarn add typescript zod express-zod-api

cat <<'END' > index.ts
import { z } from 'zod';
import { defaultEndpointsFactory } from 'express-zod-api';

defaultEndpointsFactory.build( {
  method: 'get',
  input: z.object({}),
  output: z.object({}),
  async handler() {},
} );
END

./node_modules/.bin/tsc ./index.ts
```

This is the error that is produced:

```
repro.ts:6:3 - error TS2322: Type 'ZodObject<{}, "strip", ZodTypeAny, {}, {}>' is not assignable to type 'IOSchema<any>'.
  Type 'ZodObject<{}, "strip", ZodTypeAny, {}, {}>' is not assignable to type 'ZodObject<any, any, ZodTypeAny, { [x: string]: any; }, { [x: string]: any; }>'.
    Types have separate declarations of a private property '_cached'.

6   input: z.object({}),
    ~~~~~
```

This happens because locking the version of `zod` means that unless every package in the project that uses zod also locks it to the exact same version, you are going to end up with a situation like this:

```sh
❯ find node_modules -type d -name zod
node_modules/zod
node_modules/express-zod-api/node_modules/zod
```

This means that the zod being used by express-zod-api is not the same as the zod being used in the rest of the app.

RobinTail:
Hello @jasonk 
Could you please clarify:
Why are installing `zod` having `express-zod-api` that exports `z`?

It's made this way in order to extend zod with some proprietary schemas under the same namespace, like `z.file()` etc

RobinTail:
Any updates, @jasonk ?

RobinTail:
@jasonk , consider testing v10 should fix this issue.
More info here: #870

PR #584, opened 2022-09-05
Switching to `tsup`

RobinTail:
No more magical moves with file extensions for ESM.
Using the fast modern tool for it.

PR #850, opened 2023-03-04
Detaching `ez` namespace for proprietary schemas

RobinTail:
This may help with #584 and #822


src/index.ts
```diff
@@ -27,7 +27,8 @@ export { withMeta } from "./metadata";
 export { testEndpoint } from "./mock";
 export { Client } from "./client";
 
-import * as z from "./extend-zod";
```
RobinTail, line 30: ⚠️ Breaking

RobinTail:
I think I'd better use a fork of zod with
- my custom schemas
- `.example()`

PR #865, opened 2023-03-08
For v10: Stop re-exporting `zod`

RobinTail:
For v10, #864

PR #866, opened 2023-03-08
For v10: Make `express` peer

RobinTail:
For the `attachRouting` case, developer may have their own version of `express` installed, which may cause collisions.
Doing similar thing as #865

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "@types/http-errors": "^2.0.0",
     "@types/node": "*",
     "compression": "1.7.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,6 +53,7 @@
   },
   "peerDependencies": {
     "@types/jest": "*",
+    "express": "^4.18.2",
```
RobinTail, line 56: ⚠️ Breaking

RobinTail: Not sure how to address `@types/express`...

PR #867, opened 2023-03-08
For v10: Making `typescript` a dev and optional peer dependency

RobinTail:
Typescript is only required for generating a client, the version may collide with the developer's local one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "openapi3-ts": "3.2.0",
     "ramda": "0.28.0",
     "triple-beam": "1.3.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -47,7 +47,6 @@
     "openapi3-ts": "3.2.0",
     "ramda": "0.28.0",
     "triple-beam": "1.3.0",
-    "typescript": "4.9.5",
```
RobinTail, line 50: ⚠️ Breaking: Typescript becomes an optional peer, needed for generating client only

PR #868, opened 2023-03-08
For v10: Test different typescript versions

RobinTail:


RobinTail:
Todo: v10: 4.9.3 should be the min supported peer version of ts

PR #908, opened 2023-03-27
OpenAPI fix: avoid `nullable` when `type` is not present

### LINKED ISSUE #900: `z.record(z.any())` generates invalid OpenAPI spec
Hi, I've run into some trouble with OpenAPI spec generation of input and output payloads. My goal is to accept an arbitrary object as input. The way I've found that I can accomplish this is with `z.record(z.any())` and that works just fine functionality wise. The issue I run into is that when I generate the docs, I get the following YAML in the OpenAPI spec that's generated:

```yaml
type: object
additionalProperties:
  format: any
  nullable: true
```

When I check the validity of this using a tool like Redocly's CLI, I get the following error:

```
The `type` field must be defined when the `nullable` field is used.

687 |   additionalProperties:
688 |     format: any
689 |     nullable: true
```

From looking at the OpenAPI spec, this is an accurate error. I was wondering if there was a known workaround/better way to handle a schema field for accepting an arbitrary object. Just using `z.any()` results in the same error as above.

If I manually edit the spec and add `type: object` to the `additionalProperties`, the spec stops throwing that error. Here's an example:

```
type: object
additionalProperties:
  format: any
  nullable: true
  type: object
```

I'm wondering if this is an issue with spec generation, or if I'm just trying to tackle this problem the wrong way. If you have any advice on how I should be using Zod to accept an arbitrary object that passes the OpenAPI spec, I'd love to hear.

Thank you!

RobinTail:
Closes #900

### DIFF
--- a/src/open-api-helpers.ts
+++ b/src/open-api-helpers.ts
@@ -691,14 +691,14 @@ export const depicters: HandlingRules<
 export const onEach: Depicter<z.ZodTypeAny, "each"> = ({
   schema,
   isResponse,
-  current,
+  prev,
 }) => {
-  if (isReferenceObject(current)) {
+  if (isReferenceObject(prev)) {
     return {};
   }
   const { description } = schema;
   const shouldAvoidParsing = schema instanceof z.ZodLazy;
-  const hasTypePropertyInDepiction = current.type !== undefined;
+  const hasTypePropertyInDepiction = prev.type !== undefined;
   const isResponseHavingCoercion = isResponse && hasCoercion(schema);
   const isActuallyNullable =
     !shouldAvoidParsing &&
--- a/src/open-api-helpers.ts
+++ b/src/open-api-helpers.ts
@@ -697,13 +697,16 @@ export const onEach: Depicter<z.ZodTypeAny, "each"> = ({
   const shouldAvoidParsing = schema instanceof z.ZodLazy;
   const hasTypePropertyInDepiction =
     isSchemaObject(current) && current.type !== undefined;
+  const isResponseHavingCoercion = isResponse && hasCoercion(schema);
+  const isActuallyNullable =
+    !shouldAvoidParsing &&
+    hasTypePropertyInDepiction &&
+    !isResponseHavingCoercion &&
+    schema.isNullable();
   const examples = shouldAvoidParsing ? [] : getExamples(schema, isResponse);
   return {
     ...(description && { description }),
-    ...(!shouldAvoidParsing &&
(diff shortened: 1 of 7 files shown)


src/open-api-helpers.ts
```diff
@@ -688,16 +688,20 @@ export const depicters: HandlingRules<
...
   const shouldAvoidParsing = schema instanceof z.ZodLazy;
+  const hasTypePropertyInDepiction =
+    isSchemaObject(current) && current.type !== undefined;
   const examples = shouldAvoidParsing ? [] : getExamples(schema, isResponse);
   return {
     ...(description && { description }),
     ...(!shouldAvoidParsing &&
+      hasTypePropertyInDepiction &&
       schema.isNullable() &&
       !(isResponse && hasCoercion(schema)) && { nullable: true }),
```
RobinTail, line 706: I wanna make it nicer

tests/unit/__snapshots__/open-api.spec.ts.snap
```diff
@@ -4037,7 +4032,6 @@ paths:
           description: GET /v1/getSomething parameter
           schema:
             format: string (preprocessed)
-            nullable: true
```
RobinTail, line 4040: should mention this in the changelog as well

src/open-api-helpers.ts
RobinTail: done

src/open-api-helpers.ts
```diff
@@ -688,18 +688,25 @@ export const depicters: HandlingRules<
...
-export const onEach: Depicter<z.ZodTypeAny, "last"> = ({
+export const onEach: Depicter<z.ZodTypeAny, "each"> = ({
   schema,
   isResponse,
+  current,
 }) => {
   const { description } = schema;
   const shouldAvoidParsing = schema instanceof z.ZodLazy;
+  const hasTypePropertyInDepiction =
+    isSchemaObject(current) && current.type !== undefined;
```
RobinTail, line 699: if it's a reference (somehow) we should not add any additional props to it

RobinTail: done

RobinTail (commented):
Need a specific test

tests/unit/__snapshots__/open-api.spec.ts.snap
RobinTail: done

RobinTail (commented):
Ready

PR #910, opened 2023-03-27
Fix: Status code for malformed body should be 400

### LINKED ISSUE #907: body-parser errors are not handled as HttpErrors
### Steps to reproduce:
- use `createServer` with default `server.jsonParser` to start the server
- send request with invalid JSON. e.g.:
```
curl --location 'http://localhost:{{port}}/whatever-path' \
--header 'Content-Type: application/json' \
--data '{
    a: "b"
}'
```

### Actual behaviour:
The server responds with a 500 error

### Expected behaviour
The server responds with a 400 error

---
One possible workaround is to pass a custom `jsonParser`:
```
import { isHttpError } from "http-errors";

...
const rawJsonParser = express.json();
const jsonParser: typeof rawJsonParser = (request, response, next) => {
  rawJsonParser(request, response, (error) => {
    if (error instanceof Error) {
      response.statusCode = isHttpError(error)
        ? error.status
        : getStatusCodeFromError(error);
      response.end(getMessageFromError(error));
    } else {
      next(error);
    }
  });
};
...
createServer(createConfig({
...
server: {
  jsonParser,
  ...
}
}), routing)
```

Note that code above will not work without the `isHttpError` check, because there is no such check inside `getStatusCodeFromError`. @RobinTail, Is it intentional?

RobinTail:
Fixes #907

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Version 10
 
+### v10.1.2
+
+- Fixed issue #907, found and reported by [@McMerph](https://github.com/McMerph).
+  - HTTP response status code in case of malformed body or other body-parser errors changed from `500` to `400`.
+
 ### v10.1.1
 
 - Fixed issue #900, found and reported by [Max Cohn](https://github.com/maxcohn).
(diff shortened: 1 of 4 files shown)


RobinTail (commented):
Generally ready

PR #930, opened 2023-04-10
Request param description fix

### LINKED ISSUE #929: Parameter descriptions are only outputted on referenced schema, making them invisible in docs
We recently upgraded to express-zod-api v10. This, in general, massively cleaned up our OpenAPI schema and made it work much better with code generation tools (thanks for that! 🙏). Still, it seems like it also introduced a new bug or potentially a new gotcha we weren't aware of.

Here's how to reproduce the problem:

1. Create an endpoint or middleware that has an input with a description:

```ts
export const testEndpoint = endpointFactory.build({
  input: z.object({
    cursor: z
      .string()
      .optional()
      .describe(
        'An optional cursor string used for pagination.' +
          ' This can be retrieved from the `next` property of the previous page response.',
      ),
  }),
  output: z.object({}),
  handler: async () => {
    return {}
  },
})
```

2. Generate the OpenAPI schema and look at the output:

```yml
parameters:
  - name: cursor
    in: query
    required: false
    description: GET /hris/employees parameter
    schema:
      $ref: "#/components/schemas/GetHrisEmployeesParameterCursor"
```

```yml
GetHrisEmployeesParameterCursor:
  type: string
  description: An optional cursor string used for pagination. This can be
    retrieved from the `next` property of the previous page response.
```

The custom description is outputted, but only on the referenced schema, not the actual parameter. On the parameter, the description is not only missing (which might cause documentation tools to pull the description from the schema), but specifically set to something generic (`GET /hris/employees parameter`).

The result is that no helpful information makes it all the way through to the docs:

image

I'm wondering: Is this expected behavior, and there's a workaround, or is this a bug introduced by the new major release? 🤔

RobinTail:
Fixes #929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### v10.1.3
 
 - Fixed issue #929, found and reported by [@shroudedcode](https://github.com/shroudedcode).
-  - Customized description of request parameters have not been taken into account when generating the documentation.
+  - Customized description of request parameters have not been depicted correctly when generating the documentation.
 
 ### v10.1.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Version 10
 
+### v10.1.3
+
+- Fixed issue #929, found and reported by [@shroudedcode](https://github.com/shroudedcode).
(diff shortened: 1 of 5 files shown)


RobinTail (commented):
Ready

PR #974, opened 2023-04-30
Integration variant

RobinTail:
Related to the discussion https://github.com/RobinTail/express-zod-api/discussions/931


src/integration.ts
```diff
@@ -37,6 +37,13 @@ interface Registry {
 
 interface IntegrationParams {
   routing: Routing;
+  /**
+   * @desc What should be generated
+   * @example "typesOnly" — types of your endpoint requests and responses (for a DIT solution)
```
RobinTail, line 42:
```suggestion
   * @example "typesOnly" — types of your endpoint requests and responses (for a DIY solution)
```

PR #975, opened 2023-05-01
Restoring dependency on `openapi3-ts`

RobinTail:
Since the import issue was finally fixed, I'm reverting #942.


RobinTail (commented):
LGTM

PR #984, opened 2023-05-06
Reproducing issue #976

### LINKED ISSUE #976: startupLogo is showing even when the config says it should not be
Hi,

This is the config I'm using, but in my debugger I see that the startupLogo property is no longer there after the createConfig() step.

```
createConfig({
      app,
      cors: true,
      startupLogo: false,
      logger: {
        level: 'debug',
        color: true,
      }
    });
```
image

Kind regards,
Kristof.

RobinTail:
Closes #976 

image

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -35,12 +35,6 @@
       "rules": {
         "import/no-extraneous-dependencies": "off"
       }
-    },
-    {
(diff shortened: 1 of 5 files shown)


RobinTail:
Not reproducible

PR #1028, opened 2023-06-28
Using `node:url` imports for core modules.

RobinTail:
See https://nodejs.org/api/esm.html#node-imports


RobinTail:
The ESLint rule is not yet published.
The corresponding issue: https://github.com/import-js/eslint-plugin-import/issues/2717

RobinTail (commented):
good

PR #1048, opened 2023-07-11
export getMeta

dev-m1-macbook:
hello!
I was trying to follow through https://ez.robintail.cz/v11.1.1/response-customization
and it seems that I can't continue due to `getMeta` not being exported.

this PR is a quick fix for that

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,7 +23,7 @@ export {
   OutputValidationError,
   InputValidationError,
 } from "./errors";
-export { withMeta } from "./metadata";
+export { getMeta, withMeta } from "./metadata";
 export { testEndpoint } from "./mock";
 export { Integration } from "./integration";
 


RobinTail:
I think you found a bug, @dev-m1-macbook .
`getMeta`, btw is internal low-level method. It's not supposed to be exposed. There is another one in common helpers — `getExamples`. That one should be both exposed and used in ResultHandler, since it also does parsing in case of the schema has a transformation in it.
I will take a look on it soon. Thank you for noticing this.

RobinTail:
Ah, ok. I checked it. There is no actual bug.
However, there is clearly should be something for getting examples within a custom Result Handler.

RobinTail (commented):
I'm going to solve it in a bit more advanced way in #1061, but I believe it will be more beneficial for a consumer.

RobinTail:
Thank you for noticing this problem, @dev-m1-macbook .
The resolution will arrive soon.

RobinTail:
@dev-m1-macbook , the feature released in v11.3.0 🚀 

Please use `getExamples({schema})`, without other options it acts the same as `getMeta(schema, "examples") || []`.
Thank you for your contribution.

PR #1060, opened 2023-07-13
Using unicorn's rule for validating core module imports.

RobinTail:
Better solution for https://github.com/RobinTail/express-zod-api/pull/1028


RobinTail (commented):
ok

PR #1099, opened 2023-08-19
`endpointTags` in the generated client.

RobinTail:
According to discussion #1079 

### The result

```ts
export const endpointTags = {
  "get /v1/user/retrieve": ["users"],
  "post /v1/user/:id": ["users"],
  "get /v1/avatar/send": ["files", "users"],
  "get /v1/avatar/stream": ["users", "files"],
  "post /v1/avatar/upload": ["files"],
};
```

### DIFF
--- a/src/integration.ts
+++ b/src/integration.ts
@@ -32,7 +32,10 @@ import { zodToTs } from "./zts";
 import { createTypeAlias, printNode } from "./zts-helpers";
 
 interface Registry {
-  [METHOD_PATH: string]: Record<"in" | "out", string> & { isJson: boolean };
+  [METHOD_PATH: string]: Record<"in" | "out", string> & {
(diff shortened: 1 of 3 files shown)


src/integration.ts
```diff
@@ -184,6 +188,25 @@ export class Integration {
...
+              f.createArrayLiteralExpression(
+                this.registry[methodPath].tags.map((tag) =>
+                  f.createStringLiteral(tag),
+                ),
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
```
RobinTail, line 208: I can not describe how I hate this API for such nestings :)

RobinTail: Maybe I should refactor this file to make it more readable.

miki725:
this is great! thank you

PR #1115, opened 2023-08-24
Disabling the legacy mode.

RobinTail:
Due to the following issue:
https://github.com/egoist/tsup/issues/938
which does not seem to be fixed soon.

I decided to drop the legacy mode.


RobinTail (commented):
ready

RobinTail:
That's even better

PR #1116, opened 2023-08-24
Readme: explainting the well-known issue with coerce.

### LINKED ISSUE #759: coerce and transform not having proper types in output
Typescript reports an error when using z.coerce or .transform in output.

Additionally generating an OpenAPI spec fails at runtime when using z.coere.bigint

**Minimal reproducible example** 
```ts
import { defaultEndpointsFactory, z } from "express-zod-api";

const testEndpoint = defaultEndpointsFactory.build({
  method: "get",
  input: z.object({}),
  output: z.object({
    name: z.coerce.string(),
    id: z.coerce.bigint().
  }),
  handler: async () => {
    return { name: 1, id: "1" };  // <-  Type 'number' is not assignable to type 'string'
  },
});
```

RobinTail:
Closes #759 

It does not fix it, but rather document it as a caveat and a well-known issue.


RobinTail (commented):
ok

PR #1133, opened 2023-08-30
Version in logo

RobinTail:


RobinTail:
Something is different for the `trim()` + `chalk` in CI. Not sure why, but I adjusted the test to work on all environments.

RobinTail:
don't like the side effects of this

PR #1147, opened 2023-09-06
Improving types: avoiding `any`

RobinTail:
> The key difference between unknown and any is that any allows you to perform any operation on a value without any type checking, whereas unknown requires you to perform type checking before performing any operations on the value.


RobinTail:
I'm gonna do it more precisely

PR #1181, opened 2023-09-29
Feat: ability to customize the endpoint operationId in documentation

john-schmitz:
Hi!
First of all, thank you for your effort and time in this project!

I wrote a small change in this Pull Request so we can override the generated operationId for each endpoint, while maintaining the previous behavior.

This is not a final implementation; I actually would love some pointer into how to implement this correctly;
I wrote this so we can discuss if the change is valid, at work we are using some libraries that use the operaitonId inside the swagger definition, so being able to predict how the endpoint operationId is named would help us.

Again, thank you for your time.


src/documentation.ts
```diff
@@ -68,8 +68,19 @@ export class Documentation extends OpenApiBuilder {
...
+  protected ensureUniqOperationId(
+    path: string,
+    method: Method,
+    endpointOperationId?: string,
+  ) {
+    if (
+      endpointOperationId &&
+      endpointOperationId in this.lastOperationIdSuffixes
+    ) {
+      throw new Error(`Duplicated operationId: "${endpointOperationId}"`);
```
RobinTail, line 80: this should be `DocumentationError`

RobinTail: (from `/src/errors`)

src/documentation.ts
```diff
@@ -68,8 +68,19 @@ export class Documentation extends OpenApiBuilder {
...
+    }
+
+    const operationId = endpointOperationId ?? makeCleanId(path, method);
```
RobinTail, line 83: I'd suggest a slightly different logic here and above:

- if defined by user
  - if duplicated
    - throw
  - return what's defined by user
- continue the regular flow as it is unchanged (generating by `makeCleanId`)

RobinTail: this is for keeping the user's choice as it is without any suffixes and avoid duplicated logic.
should be more clear

src/documentation.ts
RobinTail: DocumentationError's constructor also provides information on the endpoint caused the issue, could be useful

src/documentation.ts
```diff
@@ -115,6 +126,11 @@ export class Documentation extends OpenApiBuilder {
       _method,
     ) => {
       const method = _method as Method;
+      const operationId = this.ensureUniqOperationId(
+        path,
+        method,
+        endpoint.getOperationId(),
+      );
```
RobinTail, line 133: consider moving it below and placing before the `const operation`, just for keeping those things closer to each other.

src/documentation.ts
```diff
@@ -25,6 +23,8 @@ import {
   ensureShortDescription,
   reformatParamsInPath,
 } from "./documentation-helpers";
+import { mapLogicalContainer } from "./logical-container";
+import { Method } from "./method";
```
RobinTail, line 27: I'm not sure what's causing the change to the imports arrangement in several files of this PR, but if it's possible, please reduce the `diff` by restoring the original order of import (except it's not causing ESLint errors)

RobinTail (changes requested):
Thank you for the contribution, @john-schmitz .
The PR looks good.
A couple of changes requested — let me know if they are unclear.

Could you please also clarify, why it might be important to have the operation id specified manually? I think forget where it's displayed or involved.

Thank you

RobinTail:
@john-schmitz ,

perhaps utilizing the `Endpoint::getOperationId()` could also be useful in the following place (not sure yet):

https://github.com/RobinTail/express-zod-api/blob/fa47a91403ec7b70ec86c668c28dbab977e06b7d/src/integration.ts#L97-L98

src/endpoint.ts
```diff
@@ -122,6 +124,7 @@ export class Endpoint<
   };
   protected readonly scopes: SCO[] = [];
   protected readonly tags: TAG[] = [];
+  protected readonly operationId?: string;
```
RobinTail, line 127: there is a conflict due to the recent changes made in #1189 .
Please sync the `master` branch in your fork and update your PR.
The following property please rename the following way (entirely private):

```suggestion
  readonly #operationId?: string;
```

john-schmitz:
> @john-schmitz ,
> 
> perhaps utilizing the `Endpoint::getOperationId()` could also be useful in the following place (not sure yet):
> 
> https://github.com/RobinTail/express-zod-api/blob/fa47a91403ec7b70ec86c668c28dbab977e06b7d/src/integration.ts#L97-L98

@RobinTail Sorry, I did not understand what you meant by this. Could you please expand it?

src/documentation.ts
john-schmitz: It's because I had `"source.organizeImports": true` set for my typescript files in vscode.

Sorry for the confusion.

john-schmitz:
> Thank you for the contribution, @john-schmitz . The PR looks good. A couple of changes requested — let me know if they are unclear.
> 
> Could you please also clarify, why it might be important to have the operation id specified manually? I think forget where it's displayed or involved.
> 
> Thank you

To expand on it, we export the API definition so multiple clients in different languages can consume our API using generated client code. 
Consider this endpoint: `GET /cake/{cakeId}`. The generated operation ID would be something like: `GetCakeCakeId.`

For more complicated paths, it gets confusing when referencing a specific endpoint in our daily conversations and when consuming the generated client code for the API.

Its easier to write `api.GetCakeById` than `api.GetCakeCakeId` 

So, we prefer to reference specific endpoints by the operation id.

@RobinTail

john-schmitz:
Anyway, I addressed most of your comments; if I missed something, please let me know!

RobinTail:
> Sorry, I did not understand what you meant by this. Could you please expand it?

I meant that there is a feature of generating the client types (Integration), which is using the same approach for naming the IO types of endpoints. I assumed that the featured customization can also be utilized there. But it can be done later.

RobinTail:
🚀 12.3.0, @john-schmitz

PR #1183, opened 2023-09-30
Feat: headers as input source

### LINKED ISSUE #1180: Specifying input headers
I can´t find any way to specify variables in the request header the same way we can do it with params or query.
InputSources in createConfig doesn´t seem to support it like this:

`inputSources:{
      get: ["query", "params", "headers"],
    }`

And even if it´s possible to to it with the security prop on middleware like this:
`
security: {
    // this information is optional and used for generating documentation
    and: [
      { type: "input", name: "key" },
      { type: "header", name: "token" },
    ],
  },
`

I really don´t want this header to be specified as "security"

Is there any way to do this?

RobinTail:
Closes #1180 

Limitations:
- Only custom headers (starting with `x-`)
- Headers are lower case
- Duplicated headers are concatenated through `, ` — [Node.js behavior](https://nodejs.org/api/http.html#messageheaders)


src/config-type.ts
RobinTail, line 1: add a complete test for Documentation and Integration

RobinTail (commented):
ok

PR #1189, opened 2023-09-30
Fixing the bundle exports to allow consumer's declarations

### LINKED ISSUE #1182: TypeScript errors TS4023, TS4094
I recently upgraded to the latest epxress-zod-api version from 9.*, adjusted the codebase and at the end got a bunch of TS4023, TS4094 errors. I know I'm not the first one to open this issue and I know that you recommend to disable declarations in `tsconfig.json`:

> If facing Typescript errors TS4023 or TS4094, ensure disabling declaration option in your tsconfig.json.

However in my case declarations cannot be disabled. Is there a way to make it work with declarations enabled or my only option is to stay on 9.* version?

RobinTail:
Fixes #1182

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-export { createConfig, LoggerConfig } from "./config-type";
+export { createConfig, CommonConfig, LoggerConfig } from "./config-type";
 export { AbstractEndpoint } from "./endpoint";
 export { Method } from "./method";
 export {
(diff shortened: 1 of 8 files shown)


tests/unit/__snapshots__/index.spec.ts.snap
RobinTail, line 1: requires `tsd` tests to protect against occasional breaking changes

tests/unit/endpoints-factory.spec.ts
RobinTail, line 1: there is no more way to access really private properties.
But the benefit of it is that they should no longer appear in type of `Endpoint` instances

RobinTail:
🚀 12.1.0-beta1
✅ QA passed

src/index.ts
RobinTail, line 1: consider renames with underscore

RobinTail:
🚀 12.1.0-beta2
✅ QA passed

RobinTail (commented):
confirmed by the issue reported that this fix works.
ready.

PR #1194, opened 2023-10-02
Feat: the `operationId` provider

RobinTail:
An addition to #1181 

For endpoints serving several methods.

@john-schmitz , I realized that endpoints may serve several methods, while Operation ID should be unique and is assigned to each method of endpoint. So I made this feature offering an ability to specify the operation ID provider based on a method being described by the documentation generator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Version 12
 
+### v12.4.0
+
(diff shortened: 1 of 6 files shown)


john-schmitz:
Nice catch!
Honestly I don't have any strong opinions on the matter or how to solve the issue.

RobinTail (commented):
ready

PR #1196, opened 2023-10-02
Drop `DependsOnMethodError`

RobinTail:
Leaving only the programmatic check for the assignments within `DependsOnMethod`.
Moving that check to the Routing Walker.


src/index.ts
```diff
@@ -27,7 +27,6 @@ export { createServer, attachRouting } from "./server";
 export { Documentation } from "./documentation";
 export {
   DocumentationError,
-  DependsOnMethodError,
```
RobinTail, line 30: ⚠️ breaking, removing public class

src/depends-on-method.ts
```diff
@@ -1,37 +1,8 @@
...
-              `    // or method: '${key}'\n` +
-              "    ...\n" +
-              "  })\n" +
-              "});\n",
-          );
-        }
-      }
-    });
-  }
+    public readonly endpoints: Partial<Record<Method, AbstractEndpoint>>,
```
RobinTail, line 6: ⚠️ potentially breaking: renaming the public property

RobinTail (commented):
ok

PR #1198, opened 2023-10-02
Dropping the `MethodDefinition` and corresponding type argument

RobinTail:
Due to #1196 
There is actually no need to maintain the type parameter having `Method` on `Endpoint` type.
It was only required for that reason, but still it was not enough, therefore all that is replaced by a simple programmatic check in the PR mentioned above.


RobinTail (commented):
ok

PR #1199, opened 2023-10-02
Move syntax sugar

RobinTail:
I also want to eliminate the repeated types in `BuildProps`, `EndpointsProps`, as well as to move the handling variations like `method | methods` (`scope | scopes`, ...) into the Factory. So the Factory should handle all that and address the Endpoint creation with more clarity.


RobinTail (commented):
ok

PR #1214, opened 2023-10-07
Ref: `LogicalContainer`

RobinTail:
- reducing usage of `any`
- reducing conditional nesting (flattening syntax) by adding a couple more local helpers


RobinTail (commented):
nice

PR #1229, opened 2023-10-11
Better implementation for `compression` and `express-fileupload`

RobinTail:
This improve readability and makes it easier to change the implementation regarding those dependencies in the future


RobinTail (commented):
looks better

PR #1231, opened 2023-10-11
Removing redundant `isLoggerConfig` helper.

RobinTail:
This method was used to ease testing.
However, currently I know how to make it better.


RobinTail (commented):
nice

PR #1257, opened 2023-10-27
Fixing tests

RobinTail:
- no `forceExit`
  - merging client test into example test (system)
  - no detecting open handles


RobinTail:
targeting #1256

PR #1258, opened 2023-10-27
Feat: Graceful shutdown

RobinTail:
https://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html


RobinTail:
It's better to use https://github.com/gajus/http-terminator

PR #1263, opened 2023-10-28
Tests: refactoring and improving performance

RobinTail:
- `forceExit` solved issue on two different system test files running example. No needed when combined.
- `detectOpenHandles` was needed due to `forceExit`. No needed when removed.
- Integration and ESM tests should not use `yarn` since sub-sub-process killing outcome is undetectable in Linux.


tests/system/example.spec.ts
```diff
@@ -318,4 +325,40 @@ describe("Example", () => {
...
+          body: method === "get" ? undefined : JSON.stringify(params),
+        });
+        if (`${method} ${path}` in jsonEndpoints) {
+          return response.json();
+        }
+        return response.text();
+      };
+
+    const client = new ExpressZodAPIClient(
+      createDefaultImplementation(`http://localhost:${config.server.listen}`),
```
RobinTail, line 350:
```suggestion
      createDefaultImplementation(`http://localhost:8090`),
```

tests/system/example.spec.ts
```diff
@@ -1,4 +1,11 @@
 import { ChildProcessWithoutNullStreams, spawn } from "node:child_process";
+import { expectType } from "tsd";
+import { config } from "../../example/config";
```
RobinTail, line 3:
```suggestion
```

RobinTail (commented):
ok

PR #1281, opened 2023-11-05
Feat: Support raw data in requests

RobinTail:
According to discussion #869 

### The idea

- Adding `raw` option in config,
- Making additional custom middleware that places the `req.body` buffer into `req.body.raw`, so endpoints can rely on `raw` input prop,
- reusing existing `ez.file` schema
- equipping it with `.buffer()` method to change its type from `string` to `Buffer`,
- adding custom handling in case of using this schema in inputs


example/example.client.ts
```diff
@@ -92,13 +92,32 @@ type PostV1AvatarUploadResponse =
       };
     };
 
+type PostV1AvatarRawInput = {
+  raw: any;
+};
```
RobinTail, line 97: this should be fixed

example/example.client.ts
```diff
@@ -92,13 +92,32 @@ type PostV1AvatarUploadResponse =
       };
     };
 
+type PostV1AvatarRawInput = {
+  raw: string;
+};
```
RobinTail, line 97: this should be "string | Buffer" for the whole type instead

src/file-schema.ts
```diff
@@ -12,74 +12,99 @@ import { ErrMessage, errToObj } from "./common-helpers";
...
+  get isBuffer() {
+    return Buffer.isBuffer(this._def.type);
   }
 
   static create = () =>
-    new ZodFile({
-      checks: [],
-      typeName: zodFileKind,
-    });
+    new ZodFile<string | Buffer>({ typeName: zodFileKind, type: "" });
```
RobinTail, line 109: Perhaps it's better to use `string` as default for backward compatibility first

RobinTail: done

example/example.client.ts
```diff
@@ -92,13 +92,30 @@ type PostV1AvatarUploadResponse =
       };
     };
 
+type PostV1AvatarRawInput = string;
```
RobinTail, line 95: perhaps `string | Buffer`

RobinTail: taken info account

example/endpoints/accept-raw.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ez } from "../../src";
+import { taggedEndpointsFactory } from "../factories";
+
+export const rawAcceptingEndpoint = taggedEndpointsFactory.build({
+  method: "post",
+  tag: "files",
+  input: z.object({ raw: ez.file().buffer() }), // requires to enable raw option in server config
+  output: z.object({ length: z.number() }),
+  handler: async ({ input: { raw } }) => ({
+    length: raw.length,
```
RobinTail, line 11: Can I make it Plug-n-Play, without a need of using `raw`?

RobinTail: no

RobinTail: But I can make shorthands!

src/common-helpers.ts
```diff
@@ -199,33 +198,59 @@ export const hasTopLevelTransformingEffect = (schema: IOSchema): boolean => {
   return false; // ZodObject left
 };
 
-export const hasUpload = (schema: z.ZodTypeAny): boolean => {
-  if (schema instanceof ZodUpload) {
+export const hasNestedSchema = ({
```
RobinTail, line 201: perhaps it should use SchemaWalker

RobinTail: maybe next time

src/config-type.ts
```diff
@@ -50,6 +51,8 @@ export interface ServerConfig {
      * @default false
      */
     compression?: boolean | CompressionOptions;
+    /** @desc Enables parsing certain request types as raw binary data and placing it into request.body.raw property */
+    raw?: boolean | RawOptions;
```
RobinTail, line 55: perhaps I should make it `rawParser?: RequestHandler`, similar to jsonParser

src/server.ts
```diff
@@ -105,6 +105,15 @@ export const createServer = (
...
     );
   }
+  if (config.server.rawParser) {
+    app.use(config.server.rawParser);
+    app.use((req, {}, next) => {
+      if (Buffer.isBuffer(req.body)) {
+        req.body = { raw: req.body };
+      }
+      next();
+    });
```
RobinTail, line 115: perhaps I could combine them

RobinTail: maybe later, doesn't look neat

tests/express-mock.ts
```diff
@@ -3,6 +3,7 @@
 const expressJsonMock = jest.fn();
 const compressionMock = jest.fn();
 const fileUploadMock = jest.fn();
+
```
RobinTail, line 6: less diff
```suggestion
```

PR #1300, opened 2023-11-08
Fixed warnings in tests and replacing deprecated methods of `jest`

RobinTail:
In accordance with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67143

There are also some warnings fixed in tests, that are reported by IDE, like unused variables.


RobinTail (commented):
ready.
it's not urgent, but will be required for Jest 30, I presume

PR #1303, opened 2023-11-12
Testing Node 21

RobinTail:
Testing against Node 21 is for development purposes only.
Odd versions of Node are not supposed to be used on production.
Ensuring the support of Node 21 should ease the future support of Node 22 that will be production ready.

In case you need to test the compatibility of your App running `express-zod-api` against Node 21, use `--ignore-engines`  flag during installation.

https://nodejs.org/en/about/previous-releases

UPD:

- Will be replaced by #1617


RobinTail:
Closing in favour of https://github.com/RobinTail/express-zod-api/pull/1617

PR #1301, opened 2023-11-10
`v15` is for Vika

RobinTail:
### The dedication

image

Vika was 24 years young transgender woman when her boyfriend (a policeman, by the way) murdered her. Moreover, in order to confuse the tracks and create the impression that the murdered woman was alive, the policeman cut off her finger, unlocked her mobile phone and sent messages to her friends. On the evening of October 13, 2020, he was detained, confessed and showed where he buried the girl’s body.

https://ngs.ru/text/criminal/2022/09/29/71695301/

Transgender women suffer too frequently from transphobic violence and cruelty, being the less protected social group. I'd like to raise an awareness of this problem. Humans should be creators — not killers. Protect transgender women.

Find out more about Vika's story from her interview: https://ngs.ru/text/relations/2019/06/25/66137377/

----------------------

### Version 15

... description coming soon ...

- #1255 
- #1280 
- #1314 
- #1335 
  - supporting any testing framework having function mocking method 
- #1343 
- #1342 
- #1347

PR #1321, opened 2023-11-20
Ref: Async peers instead of changing config type

RobinTail:
There is a thing that I don't like in #1301 — the direction where config goes.
In my initial concept, I wanted to achieve the simplicity of config, being object of primitive properties.
Here I'd like to explore another direction by restoring back `compression` and `upload` as `bool | Options`, but instead — trying to import the corresponding peers dynamically.
For that, however, `createServer` has to become `async`, which is also breaking, but I believe, might be much less breaking, and the outcome of that should be nicer.


RobinTail (commented):
looks fine.
And it seems to be working good.
I like it this way, so the config will remain simpler.

RobinTail:
Restored documentation and adjusted the changelog.
LGTM

PR #1335, opened 2023-11-24
Augmentation of mocking function type

RobinTail:
- alternative loading of testing framework peer;
- featuring the support of even more testing frameworks


RobinTail (commented):
I like this solution

RobinTail:
`await import` does not well in jest

> You need to run with a version of node that supports ES Modules in the VM API. See https://jestjs.io/docs/ecmascript-modules

RobinTail:
This worked well for `jest`

RobinTail:
`vitest` also works good

src/common-helpers.ts
```diff
@@ -334,3 +334,25 @@ export const loadPeer = async <T>(
     throw new MissingPeerError(moduleName);
   }
 };
+
+export const loadAltPeer = async <T>(
```
RobinTail, line 338: this is probably not needed, unnecessary complication

RobinTail: but it actually helps to test what can not be tested

RobinTail (commented):
Now I like it. QAed locally via `file://` reference

RobinTail:
✅ beta4 QA passed with both jest and vitest

PR #1342, opened 2023-12-01
Ref: common starter for all servers

RobinTail:
In v15 `createServer` becomes async (returns Promise).
~~I'd like to use that advantage to await for actual listening started before fulfilling the Promise.~~

After some considerations I decided to postpone that feature and limit the PR only to refactoring with a common starter.


RobinTail (commented):
ok

PR #1343, opened 2023-12-01
Fix: supporting primitive values for logging metadata

RobinTail:
```
logger.info("Listening", 8090)
```

will work in v15


RobinTail (commented):
ok, finally

PR #1347, opened 2023-12-01
Restoring `createLogger` in v15

RobinTail:
I decided to restore `createLogger` exposition sync, but with `winston` argument.
Thus, all async `loadPeer` are concentrated in `server.ts` file and there are less breaking changes.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -35,7 +35,7 @@ export const isSimplifiedWinstonConfig = (
 /**
  * @desc a helper for creating a winston logger easier
  * @requires winston
- * @example createWinstonLogger({ winston, level: "debug", color: true })
+ * @example createLogger({ winston, level: "debug", color: true })
  * */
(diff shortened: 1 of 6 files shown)


src/logger.ts
```diff
@@ -34,20 +33,21 @@ export const isSimplifiedWinstonConfig = (
   ["silent", "warn", "debug"].includes(subject.level);
 
 /**
- * @desc an internal helper for creating a winston logger easier
+ * @desc a helper for creating a winston logger easier
  * @requires winston
- * @example await createLogger({ level: "debug", color: true })
+ * @example createWinstonLogger({ winston, level: "debug", color: true })
```
RobinTail, line 38:
```suggestion
 * @example createLogger({ winston, level: "debug", color: true })
```

RobinTail (commented):
ok

PR #1351, opened 2023-12-03
Experiment: workspaces

RobinTail:
The environment became complex enough to consider such approach.
It may help to organize different sections accordingly.
Besides that, example can refer to the distribution name instead of sources.

https://classic.yarnpkg.com/lang/en/docs/workspaces/
https://docs.npmjs.com/cli/v10/using-npm/workspaces


RobinTail:
Got it. I just need to remove example referring tests from the tests of those two classes, or split.

tools/license.ts
```diff
@@ -1,5 +1,5 @@
 import { writeFileSync } from "node:fs";
-import manifest from "../package.json";
+import manifest from "../express-zod-api/package.json";
```
RobinTail, line 2: this needs either top level tsconfig or reconsideration

RobinTail:
Symlink for Readme kinda work on the root — GitHub can render the content.
However, the disadvantage of it is a confusion with relative links, to changelog for example.
Better to find another solution.

example/example.swagger.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Example API
-  version: 15.0.0
+  version: 0.0.0
```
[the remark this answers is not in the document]
RobinTail: fixed

RobinTail:
The problem with those workspaces is that somehow `@tsd/typescript` interferes with the globals of regular `typescript`, even in workspaces that do not use `tsd`. Despite the advantage of sharing dependencies between the workspaces, it all mixes up and pollutes the globals everywhere, affecting the development experience by messing up the types, including `jest` and `vitest` globals too.
I could not find a pretty solution for that problem yet.

tsconfig.json
```diff
@@ -5,6 +5,5 @@
     "noImplicitOverride": true,
     "strictNullChecks": true,
     "resolveJsonModule": true
-  },
-  "exclude": ["tests/integration", "tests/esm", "tests/vitest"]
```
RobinTail, line 9: probably the reason for that problem is that I made top level tsconfig too inclusive

RobinTail: nope, tried this, didn't help

PR #1355, opened 2023-12-05
Ref: Stable `package.json` for integration tests

RobinTail:
Based on my findings in #1351

### DIFF
--- a/tests/integration/package.json
+++ b/tests/integration/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "express-zod-api-integration-test",
+  "private": true,
+  "version": "0.0.0",
+  "dependencies": {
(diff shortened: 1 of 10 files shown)


tests/integration/package.json
```diff
@@ -8,5 +8,8 @@
     "http-errors": "^2",
     "typescript": "^5",
     "zod": "^3"
+  },
+  "devDependencies": {
+    "@types/express": "^4"
```
RobinTail, line 13: this is for issue 952 test

RobinTail: it should move to a dedicated env I think

RobinTail (commented):
ok

Issue #952, opened 2023-04-21
Insufficient exports


RobinTail:
_Originally posted by @McMerph in https://github.com/RobinTail/express-zod-api/discussions/951_

```
TS4023: Exported variable 'getFileStreamingEndpointsFactory' has or is using name 'ZodFileDef' from external module "{{some path}}/node_modules/.pnpm/express-zod-api@10.0.0_n5zqmeauuo4jduap7lnbjtm454/node_modules/express-zod-api/dist/index" but cannot be named.
```

since 10.0.0-beta1

RobinTail:
Not fixed by #953 ¯\_(ツ)_/¯

PR #1356, opened 2023-12-05
Ref: dedicated environment for issue #952 test

RobinTail:
Also improving the performance by using `link:` in package files of tests
https://classic.yarnpkg.com/en/docs/cli/add


RobinTail (commented):
nice!

PR #1363, opened 2023-12-06
Ref: single dynamic tests making tool.

RobinTail:
Also moving tests' dependecies installation into test commands


RobinTail (commented):
ok

PR #1369, opened 2023-12-10
Switching to `vitest`

RobinTail:
Working on compatibility support of vitest in #1301 I was surprised by how good it works outta the box.
Meanwhile, I'm not satisfied by the ESM support in jest, which is probably not going to be better [in Jest 30](https://github.com/jestjs/jest/milestone/15).
So I'm exploring the possibility to switch to vitest instead.
Compatibility with jest will remain.

Also
- I don't like jest globals
- I don't like that jest changes the behaviour of `node:assert` (asserting with message:CustomError) against its documentation.


package.json
```diff
@@ -81,7 +81,6 @@
     "express": "^4.18.2",
     "express-fileupload": "^1.4.3",
     "http-errors": "^2.0.0",
-    "jest": ">=28 <30",
```
RobinTail, line 84: oops, that's wrong

tests/compat/jest.spec.ts
```diff
@@ -18,6 +17,7 @@ describe("Vitest compatibility test", () => {
       const { responseMock } = await testEndpoint({
         endpoint,
         requestProps: { method: "POST", body: { n: 123 } },
+        fnMethod: undefined,
```
RobinTail, line 20: wrong

tests/unit/server.spec.ts
```diff
@@ -21,17 +21,18 @@ import {
   createServer,
   defaultResultHandler,
 } from "../../src";
-import { AppConfig, ServerConfig } from "../../src/config-type";
+import { AppConfig, ServerConfig } from "../../src";
```
RobinTail, line 24: can combine

yarn.lock
RobinTail, line 1: rollup and esbuild seem to be duplicated.
@jest/schemas remains since vitest relies on it.

RobinTail:
I like it.
This can also enable me to use latest versions if `chalk` and `has-ansi` that are ESM only.

RobinTail:
### Performance test

| Test | Jest (master) | Vitest (feat) | Diff |
|--------|--------|--------|--------|
| Locally | 6.8-8.2 | 11.5 - 12.2s | +4.5s |

Not good, maybe it's due to reporter

RobinTail:
9.95s locally was the best result of Vitest.
Tried all available performance improvement features.

RobinTail:
Related issue: https://github.com/vitest-dev/vitest/issues/579

RobinTail:
### Performance test

| Test | Jest (master) | Vitest (feat) | Diff |
|--------|--------|--------|--------|
| Locally | ~7.5s | ~7.6s | 🆗  |

Good now.

RobinTail (commented):
alright, so far it is as much as I can do

PR #1375, opened 2023-12-11
Switching to ESM-first dual package

RobinTail:
Dual package (having the CJS version included) remains


package.json
```diff
@@ -65,7 +65,7 @@
     "*.md"
   ],
   "engines": {
-    "node": "^18.0.0 || ^20.1.0"
+    "node": "^18.6.0 || ^20.1.0"
```
RobinTail, line 68: ⚠️ Breaking
Unfortunately, can not guarantee the operation on 18.0.0.

RobinTail:
So far the integration/compatibility tests are preserved as they are.

| Test | Running |
|--------|--------|
| Compatibility | CJS |
| Integration | CJS |
| ESM | ESM | 

Perhaps, there should be some swap in naming.

example/generate-open-api-schema.ts
```diff
@@ -1,10 +1,12 @@
-import { writeFileSync } from "node:fs";
+import { writeFile } from "node:fs/promises";
 import { Documentation } from "../src";
 import { config } from "./config";
 import { routing } from "./routing";
-import manifest from "../package.json";
 
-writeFileSync(
+const manifest = (await import("../package.json", { assert: { type: "json" } }))
+  .default;
```
RobinTail, line 7: in ESM it requires assertion, but this way was only one that worked.

```ts
import manifest from "../package.json" assert { type: "json" };
```

that led to 

```
TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///../package.json" needs an import assertion of type "json"
```

which is probably an issue of `@swc-node/register/esm` loader

example/routing.ts
```diff
@@ -31,7 +31,7 @@ export const routing: Routing = {
     },
   },
   // path /public serves static files from /example/assets
-  public: new ServeStatic(join(__dirname, "assets"), {
+  public: new ServeStatic(join("example", "assets"), {
```
RobinTail, line 34: that thing does not exist in ESM. The suggested replacement is redundantly complex.

package.json
```diff
@@ -38,25 +38,25 @@
...
-        "types": "./dist/index.d.mts",
-        "default": "./dist/index.mjs"
-      },
-      "require": {
         "types": "./dist/index.d.ts",
         "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
```
RobinTail, line 59: ⚠️ _potentially_ breaking changes
(but they should not)

src/index.ts
```diff
@@ -1,19 +1,19 @@
 export { createConfig } from "./config-type";
 export { AbstractEndpoint } from "./endpoint";
-export { Method } from "./method";
+export type { Method } from "./method";
```
RobinTail, line 3: that's a weird problem

tests/unit/__snapshots__/index.spec.ts.snap
RobinTail, line 1: tracking the exported types and interfaces is no longer possible

tsconfig.json
```diff
@@ -1,10 +1,12 @@
 {
   "extends": "@tsconfig/node18/tsconfig.json",
   "compilerOptions": {
+    "module": "ESNext",
```
RobinTail, line 4: this one was requested by Typescript to support import assertions for JSON modules.
Since bundling with `tsup` it should not be an issue.

tsconfig.json
```diff
@@ -1,10 +1,12 @@
...
+    "moduleResolution": "Bundler",
     "noImplicitAny": true,
     "noImplicitOverride": true,
     "strictNullChecks": true,
     "resolveJsonModule": true
```
RobinTail, line 9: this one is problably no longer needed

package.json
RobinTail: fixed by #1376 and restored in 8e2dd056

RobinTail:
Awesome, no breaking now

example/generate-open-api-schema.ts
RobinTail: fixed by #1376 and restored in 5c02348c

tsconfig.json
RobinTail: no longer a concern after 5c02348c

tsconfig.json
RobinTail: this probably should be OK
https://www.typescriptlang.org/docs/handbook/modules/reference.html#es2015-es2020-es2022-esnext

src/index.ts
RobinTail: this issue remains even with `tsx`

tests/unit/__snapshots__/index.spec.ts.snap
RobinTail: remains with `tsx`

RobinTail: handled in 8d1ab95b

example/routing.ts
RobinTail: not an issue

RobinTail:
todo: https://github.com/RobinTail/express-zod-api/pull/1375#issuecomment-1850821269

RobinTail:
⚠️ `require` in `peer-helpers` probably needs
```ts
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);
```
https://nodejs.org/api/esm.html#differences-between-es-modules-and-commonjs

RobinTail:
> https://github.com/RobinTail/express-zod-api/pull/1375#issuecomment-1851821565

this one is actually OK.
`esbuild` removes the `require` from the ESM bundle.

tsconfig.json
RobinTail: set back to ES2022 in 2ab45441

RobinTail (commented):
looks nice

RobinTail:
🚀 15.1.0-beta1
✅ QA passed — experiment succeeded

package.json
RobinTail: confirmed to be non-breaking

package.json
```diff
@@ -25,36 +25,37 @@
...
-    "cleanup": "rm -rf tests/**/node_modules coverage tests/**/coverage",
+    "cleanup": "rm -rf tests/**/node_modules coverage",
     "precommit": "yarn lint && yarn test && yarn build && git add example/example.* LICENSE coverage.svg",
     "prepublishOnly": "yarn lint && yarn test && yarn build",
     "postversion": "git push && git push --tags",
     "install_hooks": "husky install"
   },
-  "main": "dist/index.js",
+  "type": "module",
+  "main": "dist/index.cjs",
```
RobinTail, line 47: According to examples I saw, this should now also point to ESM version, not CJS

```suggestion
  "main": "dist/index.js",
```

RobinTail:
OK. I read that `main` is actually no longer taken into account in modern Node.js when `exports` is given.
All my tests prove that it does not matter. 
Even when `require` the package from a relative path.
According to [another guide](https://www.swarmion.dev/docs/how-to-guides/migration-guides/esm-first-packages/#how), and consulting with ChatGPT, making `main` pointing to CJS version would probably make sense for some really weird backward compatibility, when, for instance, someone would try to run this on Node 10 or earlier.

RobinTail:
I decided to keep `main` pointing to CJS for weird backward compatibility cases, keeping in mind that I can change it later if needed.

package.json
RobinTail: decided to keep CJS for weird compatibility cases.
all findings tell that it really does not matter if `exports` is specified

RobinTail:
ready

PR #1376, opened 2023-12-11
Switching to `tsx` for running commands and scripts

RobinTail:
Better ESM support outta box.
Easy to use.

Press `F` to pay respect to `SWC` and `@swc-node/register`.
You did it great.


yarn.lock
```diff
@@ -1815,6 +1817,34 @@ esbuild@^0.19.2, esbuild@^0.19.3:
     "@esbuild/win32-ia32" "0.19.9"
     "@esbuild/win32-x64" "0.19.9"
 
+esbuild@~0.18.20:
```
RobinTail, line 1820: there is a duplication of esbuild, but I think it will be upgraded soon. The package looks well maintained.

PR #1385, opened 2023-12-16
Async detached Documentation

RobinTail:
I'm still exploring the ways towards selectable OAS version of Documentation and reducing dependencies that may not be required for those who do not use these features.

In this PR:
- `openapi3-ts` made an optional peer dependency to be installed only when needed the feature of generating documentation
- `new Documentation()` becomes `await createDocumentation()` exposing `{ builder, print() }`


RobinTail (commented):
basically ready

PR #906, opened 2023-03-26
Feat: OpenAPI 3.1

RobinTail:

PR #1387, opened 2023-12-18
Switching to OAS 3.1

RobinTail:
Instead of #906

In the end, after trying several approaches I decided to switch to the most modern OAS 3.1 instead of offering it as an opt-in feature.


RobinTail:
```
example/example.documentation-3.1.yaml
┌─────────┬──────────┬───────────┬───────┬────────────────────────────────────────────────────────────┐
│ (index) │   line   │ severity  │ code  │                          message                           │
├─────────┼──────────┼───────────┼───────┼────────────────────────────────────────────────────────────┤
│    0    │ '118:20' │ 'warning' │ 10069 │ 'property "example" is deprecated, use "examples" instead' │
│    1    │ '185:18' │ 'warning' │ 10069 │ 'property "example" is deprecated, use "examples" instead' │
│    2    │ '202:18' │ 'warning' │ 10069 │ 'property "example" is deprecated, use "examples" instead' │
└─────────┴──────────┴───────────┴───────┴────────────────────────────────────────────────────────────┘
```

RobinTail:
The approach is kinda good, but not perfect.
I don't like how constraints work.
Perhaps it would be better just to switch to OAS 3.1 permanently.
It's stable since November 2021.

src/documentation-helpers.ts
```diff
@@ -190,13 +187,19 @@ export const depictReadonly: Depicter<z.ZodReadonly<z.ZodTypeAny>> = ({
...
-  nullable: true,
-  ...next({ schema: schema.unwrap() }),
-});
+}) => {
+  const nested = next({ schema: schema.unwrap() });
+  return {
+    ...nested,
+    type: ["null" as Extract<SchemaObjectType, string>].concat(
+      "$ref" in nested ? [] : nested.type || [],
+    ),
```
RobinTail, line 200: this should be extracted and reused along with `onEach` handling of actually nullable stuff

src/documentation-helpers.ts
```diff
@@ -772,8 +757,12 @@ export const onEach: Depicter<z.ZodTypeAny, "each"> = ({
       });
   return {
     ...(description && { description }),
-    ...(isActuallyNullable && { nullable: true }),
-    ...(examples.length > 0 && { example: examples[0] }),
+    ...(isActuallyNullable && {
+      type: ["null" as Extract<SchemaObjectType, string>].concat(
+        prev.type || [],
+      ),
```
RobinTail, line 763: this is the same as within `depictNullable`

PR #1401, opened 2023-12-20
Feat: Multiple status codes

RobinTail:
According to discussions #1193 , #1332 and #824

I'm trying here to enable consumers to declare slightly different schemas for various HTTP status codes.
So far, it used to be only one pair, but configurable (200 and 400 by default for positive and negative accordingly).

### DIFF
--- a/src/endpoint.ts
+++ b/src/endpoint.ts
@@ -31,15 +31,28 @@ import {
 import { Security } from "./security";
 import { AbstractLogger } from "./logger";
 
-const getMimeTypesFromApiResponse = <S extends z.ZodTypeAny>(
-  subject: S | ApiResponse<S>,
-  fallback = [mimeJson],
-) => {
+interface ProcessedResponse<S> {
+  schema: S;
+  statusCodes: number[];
+  mimeTypes: string[];
+}
+
+const processApiResponse = <S extends z.ZodTypeAny>(
+  subject: S | ApiResponse<S> | ApiResponse<S>[],
+  fallback: Omit<ProcessedResponse<S>, "schema">,
+): Array<ProcessedResponse<S>> => {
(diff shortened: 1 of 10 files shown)


src/endpoint.ts
```diff
@@ -65,9 +77,10 @@ export abstract class AbstractEndpoint {
   ): string | undefined;
   public abstract getMethods(): Method[];
   public abstract getSchema(variant: IOVariant): IOSchema;
-  public abstract getSchema(variant: ResponseVariant): z.ZodTypeAny;
-  public abstract getMimeTypes(variant: MimeVariant): string[];
-  public abstract getStatusCode(variant: ResponseVariant): number;
```
RobinTail, line 70: ⚠️ potentially breaking (however, consumers are not supposed to use those methods)

src/result-handler.ts
```diff
@@ -38,8 +38,10 @@ export interface ResultHandlerDefinition<
   POS extends z.ZodTypeAny,
   NEG extends z.ZodTypeAny,
 > {
-  getPositiveResponse: (output: IOSchema) => POS | ApiResponse<POS>;
-  getNegativeResponse: () => NEG | ApiResponse<NEG>;
+  getPositiveResponse: (
+    output: IOSchema,
+  ) => POS | ApiResponse<POS> | ApiResponse<POS>[];
+  getNegativeResponse: () => NEG | ApiResponse<NEG> | ApiResponse<NEG>[];
```
RobinTail, line 44:
```
ApiResponse<NEG>[]
```

this does not work well.

RobinTail:
this solution has a tradeoff — loosing the constraint within the handler of ResultHandler in case of multiple schemas for different status codes.
And I don't like it. Wanna find a better approach.

PR #1415, opened 2023-12-24
Ref: `makeCleanId` more generic

RobinTail:
- natural order of arguments application
- more generic handling


RobinTail (commented):
amazing.
why didn't I do it earlier.

PR #1420, opened 2023-12-25
Third party licenses files

RobinTail:
Moving all third party licenses and attributions into a dedicated file


RobinTail (commented):
good

PR #1424, opened 2023-12-26
Ref: using schema for `ZodUpload` validation

RobinTail:
That custom type is actually a regular zod schema, but still with a dedicated class for handling by generators.


RobinTail (commented):
ok

PR #1425, opened 2023-12-26
Feat: Describing alternative or actual placement of the input security

RobinTail:
Resolving the long lasting todo.
According to https://swagger.io/docs/specification/authentication/api-keys/
Using the specification extensions: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specification-extensions


RobinTail (commented):
ok

PR #1431, opened 2023-12-31
Feat: Multiple status codes (2)

RobinTail:
Instead of #1401 

According to discussions https://github.com/RobinTail/express-zod-api/discussions/1193 , https://github.com/RobinTail/express-zod-api/discussions/1332 and https://github.com/RobinTail/express-zod-api/discussions/824

I'm trying here to enable consumers to declare slightly different schemas for various HTTP status codes.
So far, it used to be only one pair, but configurable (200 and 400 by default for positive and negative accordingly).


src/index.ts
```diff
@@ -49,6 +49,7 @@ export type { ZodUploadDef } from "./upload-schema";
 export type { CommonConfig, AppConfig, ServerConfig } from "./config-type";
 export type { MiddlewareDefinition } from "./middleware";
 export type { ResultHandlerDefinition } from "./result-handler";
```
RobinTail, line 51: probably need to export new type, add the case to example

src/result-handler.ts
```diff
@@ -23,7 +23,7 @@ interface ResultHandlerParams<RES> {
   logger: AbstractLogger;
 }
 
-type ResultHandler<RES> = (
+export type ResultHandler<RES> = (
```
RobinTail, line 26:
```suggestion
type ResultHandler<RES> = (
```

not needed actually

src/endpoint.ts
```diff
@@ -24,19 +24,35 @@ import { LogicalContainer, combineContainers } from "./logical-container";
...
+  defaultStatusCodes,
+} from "./result-handler";
 import { Security } from "./security";
 import { AbstractLogger } from "./logger";
 
-const getMimeTypesFromApiResponse = <S extends z.ZodTypeAny>(
-  subject: S | ApiResponse<S>,
-  fallback = [mimeJson],
-) => {
+export interface ProcessedResponse {
```
RobinTail, line 34: 1. naming
2. do I need this type?

src/index.ts
RobinTail: exposed

RobinTail: made an example including some adjustments to component naming and system test

example/routing.ts
```diff
@@ -15,8 +16,11 @@ export const routing: Routing = {
       retrieve: retrieveUserEndpoint, // path: /v1/user/retrieve
       // syntax 2: methods are defined within the route (id is the route path param by the way)
       ":id": new DependsOnMethod({
-        post: updateUserEndpoint, // the Endpoint should have at least the same method specified in .build()
+        // endpoints assigned here must support at least the same method specified in .build()
```
RobinTail, line 19:
```suggestion
        // the endpoints listed here must support at least the same method they are assigned to
```

src/endpoint.ts
RobinTail: renamed, kept.

RobinTail (commented):
kinda OK
perhaps needs a refactoring on chosen naming later

RobinTail:
✅ regression test passed

RobinTail (commented):
ok

RobinTail:
✅ QA passed (see discussions)

RobinTail (commented):
ready

PR #1448, opened 2024-01-05
Ref: `Array::some` instead of `reduceBool()`

RobinTail:
It's a day of insights today :)


RobinTail (commented):
Brilliant, Anechka :)

PR #1442, opened 2024-01-04
Proprietary schemas using native zod methods and Metadata

RobinTail:
I was dreaming about it.
Unfortunately `zod` does not provide any way to make a custom or branded or third-party schema that can be identified as one programmatically. Developer of `zod` also does not want to make any method to store metadata in schemas, instead, recommends to wrap schemas in some other structures, which is not suitable for the purposes of `express-zod-api`. There are many small inconvenient things in making custom schema classes, that I'd like to replace into native methods, and use `withMeta` wrapper for storing proprietary identifier, so the generators and walkers could still handle it.

Related issues:

```
https://github.com/colinhacks/zod/issues/1718
https://github.com/colinhacks/zod/issues/2413
https://github.com/colinhacks/zod/issues/273
https://github.com/colinhacks/zod/issues/71
https://github.com/colinhacks/zod/issues/37
```

PR I've been waiting for months to merged (programmatically distinguishable branding):

```
https://github.com/colinhacks/zod/pull/2860
```

PR #1449, opened 2024-01-05
Experiment: native depicters only

RobinTail:
Addition to #1442


RobinTail:
Nice try, but not good for `ez.file` and `ez.upload`

PR #1450, opened 2024-01-06
Experiment: `withMeta()` using `Proxy` and `Reflect`

RobinTail:
Considering to make it a bit less hacky


RobinTail:
well, it's not

RobinTail:
Actually... have one more thing to try

PR #1451, opened 2024-01-06
Experiment: Metadata in description

RobinTail:
- Instead of #1450 
- Accompanies #1442

Probably a less hacky way to store metadata.

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -46,10 +46,9 @@ export const withMeta = <T extends z.ZodTypeAny>(subject: T) =>
     get: (target, prop) => {
       if (prop === "example") {
         const setter: ExampleSetter<T> = (value) => {
-          const current = unpack(target);
-          const { examples } = current;
-          examples.push(value); // instead of concat, for handling array examples
-          return withMeta(pack(target, { ...current, examples }));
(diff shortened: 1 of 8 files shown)


RobinTail:
Major obstacle of this approach turned out to be serializing.
When example is a `Date`, it's stored as a string and deserialized incorrectly for the further examples parsing/validation.

RobinTail:
Noice! Next problem is withstanding potential calls of `.describe()` on `withMeta()` by consumer.

src/metadata.ts
RobinTail, line 1: Something is weird about this approach — too many conversions.
Need to simplify the API of this thing.

RobinTail: I'd like to simplify it, but I don't yet see how

RobinTail: Though, it works.

RobinTail, line 1: Ok. It works good now.
But it's time to reassess if it worth it mess up with a public interface of descriptions.

RobinTail:
In the end, it's an awesome implementation, but risky and harder maintainable due to increased magic behind it.
But I'm proud of you, Anya :)

PR #1457, opened 2024-01-07
Experiment: withMeta() using Proxy and Reflect.

RobinTail:
Rebased #1450 for future considerations


RobinTail:
The benefit of proxy could be a layer of abstraction without cloning the subject.
However, since other public methods create a new instances, such layer can not withstand, for example, a refinement.
Therefore, it has to keep cloning technique, which does not give much benefit including readability and maintainability, but giving a flexibility and customizations that are actually not required for this case.
In opposite, defining a property on a cloned object is way more straightforward and neatly extendible.

https://borstch.com/blog/proxies-vs-objectdefineproperty-when-to-use-which

PR #1466, opened 2024-01-09
Feat: providing `options` into ResultHandler

RobinTail:
According to the discussions #1465 and #718


RobinTail:
Decided to go with #1468 instead

PR #1471, opened 2024-01-11
Collapse empty allof declarations

arlyon:
This is perhaps a stylistic choice however I thought I'd fire a PR. I have a few layers of middleware which all have inputs by nature of requiring at least a `z.object({})`. This means I end up with many of my endpoints' types being something like (forgive pseudo-yaml)

```yaml
allOf:
  - allOf:
    - allOf:
      - type: object
        keys: ["a", "b"]
      - type: object
        keys: []
    - type: object
      keys: []
  - type: object
    keys: []
```

This PR changes the behaviour to collapse unions with empty objects such that we don't get such complicated types, as well as replacing allOfs with only a single item with just that single item.

### DIFF
--- a/src/documentation-helpers.ts
+++ b/src/documentation-helpers.ts
@@ -173,16 +173,31 @@ export const depictDiscriminatedUnion: Depicter<
   };
 };
 
-export const depictIntersection: Depicter<
-  z.ZodIntersection<z.ZodTypeAny, z.ZodTypeAny>
-> = ({
-  schema: {
-    _def: { left, right },
-  },
+export const depictIntersection: Depicter<z.ZodIntersection<z.ZodTypeAny, z.ZodTypeAny>> = ({
+  schema,
   next,
-}) => ({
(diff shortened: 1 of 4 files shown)


RobinTail:
Hello @arlyon ,

thank you for the contribution.
I understood you idea and I find it reasonable.

However, I believe it should not be limited to the case of empty declarations.
I think that every object based `allOf` could be flattened.

There is also a thing that concerns me about your implementation.
Those depicters are not supposed to dive into recursion themselves — there is a `next` method for this, provided by the schema walker. Basically, depicters should either delegate something to `next` or/and process the result of `next()`. Otherwise it creates a parallel flow of the schema tree processing, that I would like to avoid.

Therefore, I'd like to suggest you an alternative solution in #1473 .
Please check that out and let me know what you think.

arlyon:
Apologies for the slow reply, lost track of this. That change looks great! I don't have any issue with merging outputs per-endpoint so lets do it. I will close this PR. Thanks for your awesome work.

PR #1473, opened 2024-01-12
Flattening object based intersections in Documentation

RobinTail:
Alternative to #1471


src/documentation-helpers.ts
```diff
@@ -178,9 +179,37 @@ export const depictIntersection: Depicter<
...
+  const children = [left, right].map((entry) => next({ schema: entry }));
+  const safeObjects = children.filter(
+    (side): side is SchemaObject =>
+      !isReferenceObject(side) &&
+      side.type === "object" &&
+      Object.keys(side).every((key) =>
+        ["type", "properties", "required", "examples"].includes(key),
+      ),
+  );
+  if (safeObjects.length === 2) {
```
RobinTail, line 192: let's extract all that into a dedicated function

tests/unit/__snapshots__/documentation-helpers.spec.ts.snap
```diff
@@ -277,38 +277,173 @@ exports[`Documentation helpers > depictFile() > should set type:string and forma
...
           "format": "double",
           "maximum": 1.7976931348623157e+308,
           "minimum": 5e-324,
           "type": "number",
         },
       },
       "required": [
-        "two",
+        "b",
       ],
```
RobinTail, line 439: should be `a, b`, caused by:

```
flat.properties = mergeDeepRight(
      left.properties || {},
      right.properties || {},
    );
```

tests/unit/__snapshots__/documentation.spec.ts.snap
```diff
@@ -514,11 +511,8 @@ paths:
         content:
           application/json:
             schema:
-              allOf:
-                - type: object
-                  properties: {}
-                - type: object
-                  properties: {}
+              type: object
+              properties: {}
```
RobinTail, line 515: this probably can be removed as redundant later

RobinTail (commented):
it's OK now.
I'm considering making it an opt-in feature due to a possible risk of such flattening

RobinTail:
Can now be released as it is.

tests/unit/__snapshots__/documentation.spec.ts.snap
RobinTail: fixed in #1496

PR #1479, opened 2024-01-12
Ref: removing helpers that exclude

RobinTail:


src/documentation-helpers.ts
```diff
@@ -221,19 +222,27 @@ export const depictLiteral: Depicter<z.ZodLiteral<unknown>> = ({
...
-    const isOptional =
-      isResponse && hasCoercion(prop)
-        ? prop instanceof z.ZodOptional
-        : prop.isOptional();
-    return !isOptional;
-  });
+  const required = Object.keys(schema.shape)
+    .filter((key) => !excludeProps.includes(key))
+    .filter((key) => {
+      const prop = schema.shape[key];
```
RobinTail, line 231: could probably use `Object.entries` here

src/documentation-helpers.ts
```diff
@@ -62,6 +62,7 @@ export interface OpenAPIContext extends FlatObject {
   ) => ReferenceObject;
   path: string;
   method: Method;
+  excludeProps: string[];
```
RobinTail, line 65: can it be optional?

src/documentation-helpers.ts
```diff
@@ -1044,21 +1019,19 @@ export const depictRequest = ({
...
+      isResponse: false,
+      rules: depicters,
+      onEach,
+      onMissing,
+      serializer,
+      getRef,
+      makeRef,
+      path,
+      method,
+      excludeProps: pathParams,
```
RobinTail, line 1033: that's not reliable with an object depth control

PR #1518, opened 2024-01-27
Issue #1517: fixing the most negative float in depictNumber()

### LINKED ISSUE #1517: OpenApi yaml spec has wrong minimum value for doubles
# Description

If output scheme has `z.number()` minimum value in openapi sets `Number.MIN_VALUE` (> 0), but in reality floats can be negative and zero

## Expected

`z.number()` must generate openapi with correct min values

## Reproduction
1. Create handler with output schema like `{ value: z.number() }`
2. Initialize new `Documentation` object with route including handler from step.1 
3. Call `getSpecAsYaml` in `Documentation` object
4. Type description contains min specification with `minimum: 5e-324`

- Node.js version: 18.16.1
- `express-zod-api` version: 16.5.2

RobinTail:
Fixes #1517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Version 16
 
+### 16.5.3
+
+- Fixed the bug #1517 found reported by [@kotsmile](https://github.com/kotsmile):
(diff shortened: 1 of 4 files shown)


RobinTail (commented):
ready

PR #1532, opened 2024-01-30
Ref: shortening `hasNestedSchema()`

RobinTail:
Making it simpler without loosing the speed.


RobinTail:
There is a way of rewriting it using `R.either()` but it's not beneficial in terms of performance, and it does not improve readability as well.

PR #1534, opened 2024-01-31
Removing redundant condition in `combineContainers()`

RobinTail:
Excessive branch since `flattenAnds` can handle it later


RobinTail (commented):
yes

PR #1536, opened 2024-02-01
Ref: plural props

RobinTail:
Based on discussion https://github.com/RobinTail/express-zod-api/discussions/1193#discussioncomment-8014777 .

Choosing the plural version over the singular in order to imply that arrays are acceptable.


RobinTail:
Need to express `NormalizedReponse` through ApiResponse

```
Error: quick-start.ts(21,14): error TS4023: Exported variable 'helloWorldEndpoint' has or is using name 'NormalizedResponse' from external module "/home/runner/work/express-zod-api/express-zod-api/dist/index" but cannot be named.
```

RobinTail:
more harm than good

PR #1541, opened 2024-02-03
Ref: Avoiding `forEach`

RobinTail:
`for..of` is faster because it does not create a new function for each entry.


RobinTail (commented):
good

PR #1542, opened 2024-02-03
Supporting `jest` v30

RobinTail:
[Jest 30](https://github.com/jestjs/jest/milestone/15) is coming soon requiring higher Node.js version.
~~Since it's an optional peer dependency I made an arrangement to avoid breaking changes in `express-zod-api` itself.~~

Addressed to v18 having min Node version 18.18.


RobinTail:
Due to https://github.com/jestjs/jest/pull/14460 requires at least Node 18.12

RobinTail:
Or, I can make it non-breaking by testing compatibility only on certain Node versions.

RobinTail:
non-breaking now

tests/compat/package.json
```diff
@@ -13,8 +13,8 @@
     "zod": "^3"
   },
   "devDependencies": {
-    "jest": "^29.7.0",
-    "@types/jest": "^29.5.11",
+    "jest": "^30.0.0-alpha.2",
+    "@types/jest": "^29.5.12",
```
RobinTail, line 17: check stable 30 when ready

.github/workflows/node.js.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
     - name: Install Node.js ${{ matrix.node-version }}
+      id: setup-node
```
RobinTail, line 26:
```suggestion
```

PR #1545, opened 2024-02-03
Experiment: Websockets

RobinTail:
Exploring the possibilities based on the discussion: https://github.com/RobinTail/express-zod-api/discussions/991

### Concept

- Using `socket.io` library as on optional peer dependency for the best quality service including acknowledgements (I believe it's an important feature).
- Sockets is an opt-in feature by applying the new `attachSockets()` method on the `httpServer` or the `httpsServer` returned by the `createServer()` method;
- Incoming events (from a client) are handled by the entities called `Actions` having validation schemas for the incoming payload and the outgoing acknowledgement (optional), similar to `Endpoint`.
- Those schemas, in opposite to the `Endpoint`, are not `z.object()`, but `z.tuple()`.
- The returns of the Action's handler is the outgoing acknowledgement (optional), which must be the last argument received in a raw payload (a callback function, optional).
- Actions are combined into an `ActionsMap`, a dictionary associating the incoming event names with their handlers, similar to `Routing`.
- Actions are produced on an `ActionsFactory` (similar to the `EndpointsFactory`).
- That factory is a class, the constructor of which accepts the `EmissionMap` — a dictionary associating the outgoing event names with their payload validation schemas (including acknowledgements, optional).
- That makes each `Action` produced on the `ActionsFactory` aware of the available events, the payload constraints of which is ensured by the method `emit()` provided for the `Action`'s handler.
- The main `attachRouting` method also requires the `emission` for ensuring constraints on the offered features: `onConnection`, `onDisconnect` and `onAnyEvent` which are having the same type as `Action` handlers and equipped with the `emit()` method, allowing customizations.

### Next

- In case of success, the event maps can be documented according to [AsyncAPI](https://github.com/asyncapi/spec) specification.


RobinTail:
At this moment it seems that the feature should probably be a separate package.

RobinTail:
closing this PR in order to continue in a separate repo:
https://github.com/RobinTail/zod-sockets

PR #1574, opened 2024-02-20
Fix logging arrays

RobinTail:
Arrays are logged as objects with numeric keys now by the default `winston` logger.
Time to fix it.


RobinTail (commented):
ok

RobinTail (commented):
ready

PR #1585, opened 2024-02-25
Easier serialization for schemas in tests.

RobinTail:


tests/unit/__snapshots__/documentation-helpers.spec.ts.snap
```diff
@@ -1343,56 +1343,60 @@ exports[`Documentation helpers > excludeParamsFromDepiction() > should omit spec
...
-      "_type": "ZodString",
+  "_cached": null,
+  "_def": {
+    "catchall": {
+      "_def": {
+        "typeName": "ZodNever",
+      },
     },
+    "typeName": "ZodObject",
+    "unknownKeys": "strip",
```
RobinTail, line 1354: that does not show "shape"

RobinTail:
not suitable for the purpose, `inspect` does not do better as well

PR #1592, opened 2024-02-29
Feat: allowing additional options for `uploads`

### LINKED ISSUE #1590: Allow all express-fileupload configurations
Example when use limits
```ts
const serverConfig = createConfig({
  // ...
  server: {
    // ...
    upload: {
      parseNested: true,
      limits: {
        fileSize: 1024,
      },
      limitHandler: (req, res, next) => {
        return next(createHttpError(413, 'file too large'))
      }
    }
  }
}
```

throws TS2769 (No overload matches this call)

Restricted in https://github.com/RobinTail/express-zod-api/blob/master/src/config-type.ts#L78

RobinTail:
Closes #1590


src/config-type.ts
```diff
@@ -85,7 +85,17 @@ type UploadOptions = Pick<
   | "tempFileDir"
   | "debug"
   | "uploadTimeout"
->;
+  | "limits"
+> & {
+  /**
+   * @desc The error the throw when file is larger than the configured limits (handled by ResultHandler).
```
RobinTail, line 91:
```suggestion
   * @desc The error to throw when file exceeds the configured limits (handled by ResultHandler).
```

src/config-type.ts
```diff
@@ -85,7 +85,17 @@ type UploadOptions = Pick<
   | "tempFileDir"
   | "debug"
   | "uploadTimeout"
->;
+  | "limits"
+> & {
+  /**
+   * @desc The error to throw when file exceeds the configured limits (handled by ResultHandler).
+   * @see limits
+   * @override abortOnError
```
RobinTail, line 93:
```suggestion
```

RobinTail (commented):
ok

PR #1596, opened 2024-03-01
Draft solution: moving uploader into initRouting().

RobinTail:
Related to the discussion #1591


src/routing.ts
```diff
@@ -30,6 +33,12 @@ export const initRouting = ({
     routing,
     hasCors: !!config.cors,
     onEndpoint: (endpoint, path, method, siblingMethods) => {
+      // @todo this is too implicit
+      const hasUpload = endpoint.getMimeTypes("input").includes(mimeMultipart);
+      if (hasUpload && uploader) {
+        // @todo that requires error handling
+        app[method](path, uploader);
```
RobinTail, line 40: this should actually go even further, after the middlewares to enable authentication by the proprietary middlewares

RobinTail:
The problem with moving uploads handler after the middlewares is that we don't actually know what those middlewares are.
Assuming that they are only involved into authentication is still a just a particular case, not a generic solution.
It's possible that some middlewares do need files in place.

PR #1597, opened 2024-03-01
Ensure using the dedicated `next()` for `limitHandler`

RobinTail:
I found out that in some cases `req.next` may not be set and that may lead to issues.


RobinTail (commented):
nice

PR #1600, opened 2024-03-02
Fixing uncaught exception when using `limitError`

RobinTail:
The `limitHandler` of `express-fileupload` called asynchronously for a failed file before finishing the upload.
In the end `express-fileupload` calls `next()`.
All that leads to calling `next()` multiple times which leads to uncaught exception due to the way the error handling in express is organised.
The `next()` must be called once after completing the file processing.


src/server.ts
```diff
@@ -58,9 +59,11 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
         ...derivedConfig,
         abortOnLimit: false,
         parseNested: true,
-        limitHandler: limitError && (({}, {}, next) => next(limitError)),
       }),
     );
+    if (limitError) {
+      app.use(createUploadFailueHandler(limitError));
```
RobinTail, line 65: this approach can handle only `fileSize` of `limits`.

RobinTail: but that turned out to be OK because other limits do not trigger `limitHandler` —I checked it

RobinTail (commented):
good

PR #1612, opened 2024-03-10
Feat: providing `options` to Result Handler

RobinTail:
Related to discussion #1606 

Instead of https://github.com/RobinTail/express-zod-api/pull/1466 (previous attempt due to another occasion)


RobinTail (commented):
this needs a test

RobinTail (commented):
needs an update to the dataflow diagram

src/result-handler.ts
```diff
@@ -22,6 +22,8 @@ interface ResultHandlerParams<RES> {
   input: FlatObject | null;
   /** null in case of errors or failures */
   output: FlatObject | null;
+  /** can be empty: check presence of the required property using "in" operator */
+  options: FlatObject;
```
RobinTail, line 26: Current consideration: should it be consistent with `input` and `output` that are `null` in case of error?

RobinTail: or would it be excessive complication?

RobinTail: decided to keep it as it is for now

RobinTail (commented):
ok

PR #1623, opened 2024-03-16
Feat: Using the configured logger for debugging uploads

RobinTail:
I made the logger in `express-fileupload` configurable:
https://github.com/richardgirges/express-fileupload/pull/371

So now the configured logger can be used for debugging the upload instead of console.
That, however, also requires the types updated in a separate repo (waiting for that):
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69016

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -66,6 +66,10 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
         ...derivedConfig,
         abortOnLimit: false,
         parseNested: true,
+        // @todo remove ignore when the types updated:
+        // @see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69016
+        // @ts-ignore
+        logger: { log: rootLogger.debug.bind(rootLogger) },
       }),
     );
     if (limitError) {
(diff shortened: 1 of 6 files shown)


src/server.ts
```diff
@@ -66,6 +66,10 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
         ...derivedConfig,
         abortOnLimit: false,
         parseNested: true,
+        // @todo remove ignore when the types updated:
+        // @see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69016
+        // @ts-ignore
+        logger: { log: rootLogger.debug.bind(rootLogger) },
```
RobinTail, line 72: ⚠️ todo
```suggestion
        logger: { log: rootLogger.debug.bind(rootLogger) },
```

src/server.ts
RobinTail, line 1: update the `@types/express-fileupload` when the following PR merged:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69016

src/uploads-fix.d.ts
RobinTail, line 1: todo: remove this file when typed updated

src/server.ts
RobinTail: can do it independently later

src/uploads-fix.d.ts
RobinTail: can do it later independently

RobinTail (commented):
ok

PR #1633, opened 2024-03-19
Ignoring engines for `devDependencies` only

RobinTail:
That's how `yarn --production` should work but it does not.
The idea here is to reduce the application of `--ignore-engines` only to the `devDependencies`.
So that I could keep the `engines` of `express-zod-api` intact and it will reflect the requirements of the library in runtime.


RobinTail:
All good expect for compat test of jest 30, that could be restored as conditional one

PR #1654, opened 2024-03-30
Replacing Chalk with Ansis

RobinTail:
- it's faster
- it supports CJS
- it has named exports
- Similar to Chalk it now supports fallbacks from True Color

Related topic: https://github.com/webdiscus/ansis/issues/18


RobinTail (commented):
awesome!

PR #1671, opened 2024-04-07
Ref: Deasync `attachRouting()` in v18

RobinTail:
In v18 after detaching from winston, certain entities can be made sync again.


RobinTail (commented):
ok

PR #1672, opened 2024-04-07
Replacing tsx with vite-node.

RobinTail:
Considering `vite-node` as an alternative to `tsx`.
I actually like `tsx`, it works flawlessly, but since I already have `vitest`, the `vite-node` is also installed.
Therefore, using `vite-node` can reduce the number of dependencies and ecosystems involved.


RobinTail:
I'm not satisfied with its performance. It seems to be 50% slower than `tsx`.

PR #1673, opened 2024-04-07
Upgrading all dependencies using latest classic yarn (1.22.22).

RobinTail:
Found out that my local `yarn` was outdated


RobinTail (commented):
ok

PR #1677, opened 2024-04-09
Forcing NPM registry to improve `yarn` stability

RobinTail:
CI fails to often recently due to the failures of yarn's registry


RobinTail:
does not fix

Issue #1684, opened 2024-04-19
tsc error when trying to build using the getting started instructions


brainthinks:
# Description

I followed the getting started instructions from the master branch README in a new project. When running tsc on my project, I received the following:

```
$ tsc
node_modules/express-zod-api/dist/index.d.ts:1:25 - error TS2307: Cannot find module 'compression' or its corresponding type declarations.

1 import compression from 'compression';
                          ~~~~~~~~~~~~~

node_modules/express-zod-api/dist/index.d.ts:3:37 - error TS2307: Cannot find module 'express-fileupload' or its corresponding type declarations.

3 import * as express_fileupload from 'express-fileupload';
                                      ~~~~~~~~~~~~~~~~~~~~

node_modules/express-zod-api/dist/index.d.ts:4:41 - error TS2307: Cannot find module 'express-fileupload' or its corresponding type declarations.

4 import express_fileupload__default from 'express-fileupload';
                                          ~~~~~~~~~~~~~~~~~~~~

node_modules/express-zod-api/dist/index.d.ts:8:21 - error TS2307: Cannot find module 'winston' or its corresponding type declarations.

8 import Winston from 'winston';
                      ~~~~~~~~~

node_modules/express-zod-api/dist/index.d.ts:523:40 - error TS2344: Type '"body" | "headers" | "query" | "params" | "files"' does not satisfy the constraint 'keyof Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.
  Type '"files"' is not assignable to type 'keyof Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.

523 type InputSource = keyof Pick<Request, "query" | "body" | "files" | "params" | "headers">;

```

## Expected

I expected the getting started instructions to result in a successful tsc build.

## Reproduction

Copy and paste the code snippets in the getting started section of the README.

## Context

- Node.js version: 20.9.0
- `express-zod-api` version: 17.7.0

Also worth noting that i'm attempting to pass an instance of Pino rather than letting express-zod-api instantiate a winston logger for me.

brainthinks:
In order to unblock myself, I ran the following:

```
yarn add -D @types/compression @types/express-fileupload @types/winston
```

I may have missed something, and installing these types may not be the desired fix, but it's at least letting me progress.

RobinTail:
@brainthinks ,

Your dev environment tries to compile/validate `node_modules`.
[...]

PR #1696, opened 2024-04-23
add optional type dependency installation instructions

brainthinks:
Adds a sub heading in the "Installation" section for optional dependency installation to support projects that don't use `skipLibCheck`.

Addresses https://github.com/RobinTail/express-zod-api/issues/1684

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,21 @@ Ensure having the following options in your `tsconfig.json` file in order to mak
 }
 ```
 
+### Optional Dependencies
+
+If you want or need to omit the `skipLibCheck` setting recommended above, you can install the following
+optional type dependencies to ensure that `express-zod-api` doesn't fail when compiling via `tsc`:
+
+```shell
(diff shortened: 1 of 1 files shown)


brainthinks:
Ok well I now think that this PR can be disregarded / declined. I ran into a use case where I now understand why `skipLibCheck` is in the recommended tsconfig. I've been doing stateless apps for about the last 3 years, and the one I'm working on now is using RDS. I installed Drizzle ORM and immediately ran into this issue. Having seen this, it is clear to my why I shouldn't need to install all of these optional types. With express-zod-api, it was only 2 additional @types dependencies. With Drizzle, I'd have to install entire DB dependencies (since the types are included with the packages, not via DefinitelyTyped) that I'm obviously never going to use. Having to install entire mysql dependencies when i'm using postgres would be absurd.

RobinTail:
ok, no worries

PR #1707, opened 2024-04-27
Removing condition on compat test for jest in v19

RobinTail:
Always compatible now


RobinTail:
SWC core is having release trouble

```
https://github.com/swc-project/swc/issues/8898
```

RobinTail (commented):
ok now

PR #1719, opened 2024-05-01
Metadata without `withMeta()`

RobinTail:
According to the Zod author's PR, it's kinda OK to alter Zod's prototypes:

https://github.com/colinhacks/zod/pull/3445/files#diff-d8a41ab20c64f92092513f153b78e7bbf5dc929c92909fad25e03d53c7f15bb7R21-R37

I tried to avoid that when I created `withMeta()`, but since Colin is doing it themselves, this approach could be beneficial for consumers.


RobinTail:
this won't work by altering prototypes of `ZodType`, because it's not enough for `.example()` to return `this` for being a particular class instance.

src/metadata.ts
```diff
@@ -3,83 +3,86 @@ import { z } from "zod";
...
-type WithMeta<T extends z.ZodTypeAny> = T & {
-  _def: T["_def"] & Record<typeof metaProp, Metadata<T>>;
-  example: ExampleSetter<T>;
-};
+declare module "zod" {
+  interface ZodTypeDef {
+    [metaSymbol]: Metadata;
+  }
+  interface ZodType {
+    example: (example: this["_input"]) => this;
```
RobinTail, line 23: this does not work for my case

RobinTail: that way I'd have to alter prototypes of each class and I'll not do that

RobinTail:
my bad, there is probably a way...

src/metadata.ts
RobinTail: the issue was the definition.
it has to be declared as method, not property having method.

RobinTail:
The approach on altering prototypes of `ZodType` is confirmed by Colin:
https://github.com/colinhacks/zod/pull/3445#issuecomment-2091463120

The final plugin implementation is here (slightly changed):
https://github.com/colinhacks/zod/blob/90efe7fa6135119224412c7081bd12ef0bccef26/plugin/effect/src/index.ts#L21-L51

RobinTail (commented):
ok

RobinTail:
✅ QA passed

PR #1730, opened 2024-05-05
Complete `zod` plugin with proprietary brands

RobinTail:
Due to #1721, https://github.com/colinhacks/zod/pull/2860#issuecomment-2092003453, and #1719 I'm exploring the possibility to alter the behaviour of `.brand()` for storing the `brand` property as a way to distinguish the proprietary schemas in runtime.

This can replace the `proprietary()` function with a call of `.brand()`.

However, so far it turned out to be breaking because `ZodBranded` does not expose the methods of the wrapped schema, such as `.extend()`, which is used in `accept-raw.ts` endpoint for possible route params.

——

After a series of serious considerations I realized that exposing brand to consumers of `express-zod-api` could be a beneficial feature.
Runtime brand can be accessed via `._def[Symbol.for("express-zod-api")].brand`


example/endpoints/accept-raw.ts
```diff
@@ -5,9 +5,10 @@ import { taggedEndpointsFactory } from "../factories";
 export const rawAcceptingEndpoint = taggedEndpointsFactory.build({
   method: "post",
   tag: "files",
-  input: ez
-    .raw() // requires to enable rawParser option in server config
-    .extend({}), // additional inputs, route params for example, if needed
+  // requires to enable rawParser option in server config:
+  input: ez.raw({
+    /* additional inputs, route params for example, if needed */
+  }),
```
RobinTail, line 11: ⚠️ this is breaking

src/io-schema.ts
```diff
@@ -14,7 +14,8 @@ export type IOSchema<U extends z.UnknownKeysParam = any> =
   | z.ZodUnion<[IOSchema<U>, ...IOSchema<U>[]]>
   | z.ZodIntersection<IOSchema<U>, IOSchema<U>>
   | z.ZodDiscriminatedUnion<string, z.ZodObject<any, U>[]>
-  | Refined<z.ZodObject<any, U>>;
+  | Refined<z.ZodObject<any, U>>
+  | z.ZodBranded<IOSchema<U>, any>;
```
RobinTail, line 18: this is actually exclusively for `ez.raw()`, perhaps can be narrowed

example/endpoints/accept-raw.ts
RobinTail: and there is an issue with type inference of `raw`

RobinTail:
At the moment, despite the beauty of it, I'm thinking that it's not needed to be a part of consumer's API, so it can be kept as an internal method for a while.

example/endpoints/accept-raw.ts
RobinTail: UPDATE:
.extend issue was fixed in 3.23.7
https://github.com/colinhacks/zod/issues/3437

RobinTail:
UPDATE on raw / extending shape issue:
https://github.com/RobinTail/express-zod-api/pull/1730#discussion_r1594067126

RobinTail:
I realized that this can actually be a feature for those who may need it as much as I am.

example/endpoints/accept-raw.ts
RobinTail: fixed the issue

src/io-schema.ts
RobinTail: addressed and tested in a2488828

src/metadata.ts
RobinTail, line 1: I'm thinking about the idea on transforming `getMeta()` into particular methods of `ZodType` following the concept of zod plugin.

RobinTail: it does not work well for `ZodBranded`

example/endpoints/accept-raw.ts
RobinTail: migration described in 47308116

RobinTail (commented):
ok, I'm kinda proud of it

PR #1735, opened 2024-05-06
Expressing `Method` via `IRouter`

RobinTail:
This serves two purposes:

- Enforcing constraints on its actual usage in `walkRouting()`,
- Should ease adoption of the new `QUERY` method in the future, see:
  - https://github.com/expressjs/express/issues/5615
  - https://github.com/nodejs/node/issues/51562


RobinTail (commented):
ok

PR #1736, opened 2024-05-06
Fix: freezing arrays returned by public methods

RobinTail:
This is a fix that prevents changing arrays after they have been assigned upon building an Endpoint, however, it could be potentially breaking, therefore I'm addressing it to v19.


RobinTail (commented):
ok

PR #1737, opened 2024-05-06
Prefer `Array::slice()` over `Array.from()`

RobinTail:
`Array::slice()` appears to be [twice faster](https://www.measurethat.net/Benchmarks/Show/30704/0/arrayfrom-vs-arrayslice-2)


src/integration.ts
```diff
@@ -258,47 +258,47 @@ export class Integration {
...
-                ? makeInterfaceProp(quoteProp(method, path), reference)
-                : undefined;
-            })
-            .filter(
-              (entry): entry is ts.PropertySignature => entry !== undefined,
-            ),
-        ),
-      );
+      const props: ts.PropertySignature[] = [];
+      for (const [{ method, path }, entry] of this.registry) {
```
RobinTail, line 264: can I scan it once?

RobinTail: fixed in 9784436d

RobinTail (commented):
I like it

PR #1739, opened 2024-05-07
Exposing the expected request content type

RobinTail:
This should enable connecting uploader/raw parsers only to the endpoints that really need it (initRouting).
Then can revert #1733


RobinTail:
The following changes might be breaking for the type `beforeUpload` hook, because currently it receives the whole `app`, while it should be receiving the arguments natural for a middleware.
Therefore it will be addressed in another PR targeting v19.

RobinTail (commented):
ok

PR #1741, opened 2024-05-07
Feat: Selective parsers

RobinTail:
Enabled by #1739 
Addressing v19

This should connect only the parsers needed for particular endpoint.
Then can revert #1733

PR #1745, opened 2024-05-09
Automatic default raw parser

RobinTail:
Due to #1741 , the default raw parser can now be set automatically, so that the config option can now be only used for customizations.


RobinTail (commented):
ok

PR #1692, opened 2024-04-22
`v19` is for Dime

RobinTail:
image

Pebbles LaDime (Dime) Doe was a black transgender woman.

https://edition.cnn.com/2024/02/25/us/dime-doe-trial-south-carolina-federal-hate-crime/index.html
https://www.nbcnews.com/feature/nbc-out/south-carolina-death-marks-14th-black-transgender-woman-killed-u-n1040971
https://www.justice.gov/opa/pr/south-carolina-man-found-guilty-hate-crime-killing-transgender-woman-because-her-gender

Transgender women suffer too frequently from transphobic violence and cruelty, being the less protected social group. I'd like to raise an awareness of this problem. Humans should be creators — not killers. But most importantly, I want every transgender girl to have an opportunity to create applications quickly and, in general, learn to write code easily in order to receive job offers and leave dangerously transphobic territories for more favorable and civilized ones, and live happily there. Protect transgender women.

------------------------------

This version is focused on making `express-zod-api` a complete `zod` plugin. In this regard `withMeta` is removed in favor of the recommended approach on extending `zod` functionality, which opens up opportunities for new features and simplifies the consumer experience. Another improvement has been made for parsers: they are now applied selectively depending on the type of expected request.

- #1693 
- #1705 
- #1707 
- Theoretically #1631 
  - In advance #1708 
- #1726 
- #1736 
- #1741 
  - #1745 
  - #1762 
  - #1756 
  - #1766 
- #1730 
  - #1747 
  - #1748 
- #1755 
- #1760 

- Not included, but comes later: #1750 

⚠️ don't squash it, to avoid conflicts with #1750

PR #1747, opened 2024-05-09
Fix TS2527 for unique symbol (v19)

RobinTail:
https://github.com/RobinTail/express-zod-api/pull/1692#issuecomment-2102867495


RobinTail (commented):
nice!

PR #1753, opened 2024-05-11
Feat: Support async functions by `.addOptions()`

RobinTail:
I read the documentation and realized that it does not make sense for `.addOptions` just to proxy the static data.
For that purpose one could just `import` the previously assigned `const`. That's it.
What does make sense is to accept an async function by that method.

Static option will be deprecated and removed in v19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Version 18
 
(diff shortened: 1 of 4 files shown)


RobinTail (commented):
nice!

PR #1760, opened 2024-05-12
Increasing minimal versions of express and upload peers

RobinTail:
I almost forget to do an important thing for v19:

- Express should be at least 4.19.2 (lower ones are [vulnurable](https://github.com/expressjs/express/security/advisories/GHSA-rv95-896h-c2vc))
- File uploads should be at least 1.5.0 — it has customizible logger that I made and that is being used by the library


RobinTail (commented):
noice!

PR #1769, opened 2024-05-15
Header params are missing or duplicated

### LINKED ISSUE #1768: Duplicated headers in request body in generated docs (POST, PATCH, etc.)
# Description

When specifying custom headers in GET endpoints, the generated documentation is correctly built with the header field separate from the query params. In case of other HTTP Verbs (POST, PATCH, etc.), the documentation incorrectly includes the header in the request body instead of making a header field, though after specifying the `inputSources` in the `config`, request works in both cases, including the value in the request body and not in the headers OR including it in the headers and not in the request body.

## Expected

Generated documentation should correctly specify the field as a header in case of POST and other HTTP Verbs.

## Reproduction

Sample Repo:
https://github.com/boarush/express-zod-api/tree/bug/missing-headers

Have forked this repo and modified the example GET, POST and PATCH endpoints for in the user section.

## Context

- Node.js version: 20.12.2
- `express-zod-api` version: 19.1.0

Thank you for the wonderful work! Really simple working with the library!

RobinTail:
Fixes #1768 

1. Body depiction does not exclude header params (only path params);
2. Params depiction only depicts header params when `query` is enabled as input source.

### DIFF
--- a/tests/unit/__snapshots__/documentation.spec.ts.snap
+++ b/tests/unit/__snapshots__/documentation.spec.ts.snap
@@ -3866,6 +3866,13 @@ paths:
   /v1/test:
     put:
       operationId: PutV1Test
+      parameters:
+        - name: x-request-id
+          in: header
+          required: true
+          description: PUT /v1/test Parameter
+          schema:
+            type: string
       requestBody:
         description: PUT /v1/test Request body
         content:
--- a/tests/unit/__snapshots__/documentation.spec.ts.snap
+++ b/tests/unit/__snapshots__/documentation.spec.ts.snap
@@ -3791,14 +3791,8 @@ paths:
           application/json:
             schema:
               type: object
-              properties:
-                id:
-                  type: string
-                x-request-id:
-                  type: string
-              required:
-                - id
-                - x-request-id
+              properties: {}
+              required: []
       responses:
         "200":
           description: POST /v1/test Positive response
@@ -3882,11 +3876,8 @@ paths:
               properties:
                 id:
                   type: string
-                x-request-id:
-                  type: string
               required:
                 - id
-                - x-request-id
       responses:
         "200":
           description: PUT /v1/test Positive response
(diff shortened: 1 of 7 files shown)


tests/unit/__snapshots__/documentation.spec.ts.snap
```diff
@@ -3763,6 +3763,100 @@ servers:
...
+          required: true
+          description: POST /v1/test Parameter
+          schema:
+            type: string
+        - name: x-request-id
+          in: header
+          required: true
+          description: POST /v1/test Parameter
+          schema:
+            type: string
```
RobinTail, line 3787: this is OK

tests/unit/__snapshots__/documentation.spec.ts.snap
```diff
@@ -3763,6 +3763,100 @@ servers:
...
+        description: POST /v1/test Request body
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                id:
+                  type: string
+                x-request-id:
+                  type: string
```
RobinTail, line 3798: this is not OK

tests/unit/__snapshots__/documentation.spec.ts.snap
```diff
@@ -3763,6 +3763,181 @@ servers:
...
+        description: PUT /v1/test Request body
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                id:
+                  type: string
+                x-request-id:
+                  type: string
```
RobinTail, line 3879: this is also wrong. Moreover, parameters are missing because "query" is disabled.
Additional case.

RobinTail: missing header param fixed in 4845123f, but needs a refactoring

tests/unit/__snapshots__/documentation.spec.ts.snap
RobinTail: fixed

tests/unit/__snapshots__/documentation.spec.ts.snap
RobinTail: fixed

RobinTail (commented):
ok

RobinTail (commented):
Ready

PR #1778, opened 2024-05-18
Built-in logger class with `.child()` method

RobinTail:
Can do this now (without additional libraries):

```ts
import { randomUUID } from "node:crypto";
import { BuiltinLogger, createConfig } from "express-zod-api";

declare module "express-zod-api" {
  interface LoggerOverrides extends BuiltinLogger {}
}

const config = createConfig({
  logger: { level: "debug", color: true },
  childLoggerProvider: ({ parent }) =>
    parent.child({ requestId: randomUUID() }), // context
});
```

`requestId` is special and displayed after the timestamp, looks this way:

image


RobinTail (commented):
ready

PR #1786, opened 2024-05-23
Lint: addressing `any` in sources

RobinTail:
Some public interfaces changed.
Should not be an issue, but nominally it's a breaking change. Therefore, addressing to v20

### DIFF
--- a/src/testing.ts
+++ b/src/testing.ts
@@ -1,5 +1,6 @@
 import { Request, Response } from "express";
 import http from "node:http";
+import { FlatObject } from "./common-helpers";
 import { CommonConfig } from "./config-type";
 import { AbstractEndpoint } from "./endpoint";
 import { AbstractLogger, ActualLogger } from "./logger-helpers";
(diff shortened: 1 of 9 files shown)


tests/unit/server-helpers.spec.ts
```diff
@@ -94,7 +94,7 @@ describe("Server helpers", () => {
           locals: {
             [metaSymbol]: { logger: { ...rootLogger, isChild: true } },
           },
-        } as unknown as Response,
```
RobinTail, line 97: consumers should avoid doing that

src/testing.ts
```diff
@@ -118,9 +120,9 @@ interface TestEndpointProps<REQ, RES, LOG> {
 
 /** @desc Requires either jest (with @types/jest) or vitest or to specify the fnMethod option */
 export const testEndpoint = async <
-  LOG extends Record<string, any>,
-  REQ extends Record<string, any>,
-  RES extends Record<string, any>,
+  LOG extends FlatObject,
+  REQ extends FlatObject,
+  RES extends FlatObject,
```
RobinTail, line 125: ⚠️ this is potentially breaking, so that addressing to v20

RobinTail: addressed

PR #1788, opened 2024-05-24
Lint: Addressing empty object type in tests and sources

RobinTail:
https://www.totaltypescript.com/the-empty-object-type-in-typescript

### DIFF
--- a/src/common-helpers.ts
+++ b/src/common-helpers.ts
@@ -11,6 +11,7 @@ import { AuxMethod, Method } from "./method";
 import { contentTypes } from "./content-type";
 
 export type FlatObject = Record<string, unknown>;
+export type EmptyObject = Record<string, never>;
 
 const areFilesAvailable = (request: Request): boolean => {
   const contentType = request.header("content-type") || "";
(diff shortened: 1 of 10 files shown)


src/common-helpers.ts
```diff
@@ -11,6 +11,7 @@ import { AuxMethod, Method } from "./method";
 import { contentTypes } from "./content-type";
 
 export type FlatObject = Record<string, unknown>;
+export type EmptyObject = Record<string, never>;
```
RobinTail, line 14: this type does not allow assignments, but it works for reading when merged with another interface. It should not be breaking, but still addressing that change to v20, because it narrows some other publicly exposed interfaces

src/endpoints-factory.ts
```diff
@@ -1,6 +1,6 @@
 import { Request, Response } from "express";
 import { z } from "zod";
-import { FlatObject } from "./common-helpers";
+import { EmptyObject, FlatObject } from "./common-helpers";
```
RobinTail, line 3:
```suggestion
import type { EmptyObject, FlatObject } from "./common-helpers";
```

src/common-helpers.ts
RobinTail: addressed

RobinTail (commented):
ok

PR #1792, opened 2024-05-25
Middleware class

RobinTail:
https://github.com/RobinTail/express-zod-api/pull/1787#discussion_r1613867932

PR #1787, opened 2024-05-24
`v20` is for Zoey

RobinTail:
image

Zoey Alexandria, a trans woman and a voice actress has died at the age of 29 (30.04.2024).

https://www.thepinknews.com/2024/05/07/zoey-alexandria-dead-bydaylight-trans-actress-death/
https://www.them.us/story/dead-by-daylight-voice-actor-zoey-alexandria-death

--------------------

This version focuses on improving typing, consistency of methods and approaches, and simplifying the testing approach. A number of creator functions have been replaced with classes. Creating ResultHandlers and Middlewares has become more flexible. To eliminate frequent confusion, a new approach to endpoint testing has been applied: request and response mocking is completely detached from testing frameworks and works out of the box without any tricks. For the first time, an automatic migration is offered as an ESLint plugin, which itself adapts your code to this version.

- #1786 
- #1788 
- #1789 
- #1793 
- #1792 
- #1796 
- #1797 
- #1800 
- #1801 
- #1802 
- #1803 
- #1804
- #1805 
- #1847 
- Rejected: #1854

PR #1796, opened 2024-05-26
Simpler intersections

RobinTail:
After #1792 it's time to get rid of redundant complication and use regular intersection schema.
Utilizing the `EmptyObject` type introduced in #1788 for v20 #1787


RobinTail (commented):
No `bool` sheet!

PR #1802, opened 2024-06-02
Detach from `jest` and `vitest` peers

RobinTail:
Due to #1800 

Testing framework does not really matter after that since no longer using its mocks.


RobinTail (commented):
ok

PR #1847, opened 2024-06-19
Entrypoint for migration

RobinTail:
This should detach the migration helper from the actual library code and make it easier to maintain and drop if needed.


RobinTail (commented):
ok

PR #1854, opened 2024-06-23
Experiment: no dynamically loaded peers

RobinTail:
This can be a feature (minor release) in v20.
I'm looking for the opportunity to make it simpler.
Users can still upgrade sub-dependencies by running `yarn upgrade` or similar command of their package manager, or by reinstalling `express-zod-api`.


RobinTail:
Not sure about this [6e16524](https://github.com/RobinTail/express-zod-api/pull/1854/commits/6e16524ed993074a8fea39f348a5388272e9a041)

Pro:
- it's easier to use 

Cons:
- breaking change to `beforeRouting` (reduced feature)
- locking the further development of such feature as "wait before it actually starts listening"

RobinTail:
I took a look how many dependencies `prettier` installs and I don't wanna do this.
Let it be peers for a while

RobinTail:
I actually wanted the uploads to be an opt-in feature. Not everyone needs it. Also, Such things as `prettier` can also be involved by consumer, and I didn't want multiple versions to be installed.

PR #1866, opened 2024-06-29
Interface correction

RobinTail:
Found this discrepancy while trying the future zod 4 (no beta even published yet).


RobinTail (commented):
ok

PR #1869, opened 2024-06-30
Feat: top level transformations and remapping

RobinTail:
Based on the discussion #1615 

⚠️ Only transformations to another object will be allowed

Todo:

- handle output schema transformations in the documentation ✅ 
  - done by adding `ZodObject::remap()` to the distributed Zod plugin 
- more tests ✅ 
- documentation ✅


tests/unit/__snapshots__/documentation.spec.ts.snap
```diff
@@ -3683,6 +3683,80 @@ servers:
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    const: success
+                  data:
+                    format: any
```
RobinTail, line 3714: this happens because the transformation result is not a primitive type, and that's not handled

RobinTail: fixed

src/mapping-helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+    ? TT extends [infer P, V]
+      ? P
+      : never
+    : never;
+
+export type Remap<T, U extends { [P in keyof T]: V }, V extends string> = {
+  [P in U[keyof U]]: T[GetKeyByValue<U, P>];
+};
+
+export const asTuple = <A, B>(a: A, b: B): [A, B] => [a, b];
```
RobinTail, line 20: `asPair`

src/mapping-helpers.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @fileoverview Service types for Zod Runtime Plugin, mapping utils in particular.
```
RobinTail, line 2:
```suggestion
 * @fileoverview Mapping utils for Zod Runtime Plugin (remap)
```

RobinTail (commented):
reviewed

src/mapping-helpers.ts
RobinTail: replaced with R.pair

RobinTail:
✅ QA passed

PR #1878, opened 2024-07-02
Feat: Partial mapping and passthrough support for `.remap()`

RobinTail:
Addition to #1869 

Todo:
- also handle/proxy passthrough in pipe (`this._def.unknownKeys`) ✅ 
- handle explicit `undefined` ✅


RobinTail (commented):
okay :)

PR #1884, opened 2024-07-03
Recommend shallow top level transformations

RobinTail:
https://www.npmjs.com/package/camelize-ts#shallow-option


RobinTail (commented):
yes

PR #1893, opened 2024-07-08
Replacing `tsd` with `vitest` type assertions.

RobinTail:
Reducing dependencies that way.

https://vitest.dev/api/expect-typeof.html#expecttypeof

This PR also fixes a typo in JSDoc and adds missing type constraint test for ZodObject.


yarn.lock
```diff
@@ -2325,21 +2222,6 @@ jackspeak@^3.1.2:
   optionalDependencies:
     "@pkgjs/parseargs" "^0.11.0"
 
-jest-diff@^29.0.3:
```
RobinTail, line 2328: holy moly

RobinTail (commented):
good

PR #1919, opened 2024-07-18
Fixes for Node 22.5

RobinTail:


RobinTail:
Ok. It's not because of this.
https://github.com/nodejs/node/issues/53902

PR #1930, opened 2024-07-22
Better `tsup` config for CJS build of migration

RobinTail:
I found an option for what I've been doing manually.


RobinTail (commented):
ok

PR #1932, opened 2024-07-22
Preserve `node:` protocol

RobinTail:
Node protocol removal was introduced here for CJS:
https://github.com/egoist/tsup/pull/460

The target was set to `es2022` which does not explicitly address Node, therefore it was removed for ESM too.

Setting the target to `node18.18.0` enabled the protocol for ESM.

The recently added setting allows also to enable it for CJS:
https://github.com/egoist/tsup/commit/e7ced3474a559e35fc5b9cbd4cfd4aa39d44e0c0

This will certainly not work in Node below 14.18, but it should not anyway.


tsup.config.ts
```diff
@@ -8,6 +11,9 @@ const commons: Options = {
   clean: true,
   dts: true,
   minify: true,
+  target: `node${minNode.major}.${minNode.minor}.${minNode.patch}`,
+  platform: "node",
+  removeNodeProtocol: false, // @todo will be default in v9
```
RobinTail, line 16: should be redundant

PR #1969, opened 2024-08-05
Replacing ESLint `import-x` plugin with `no-restricted-imports`

RobinTail:
Maintainers can not maintain.


RobinTail (commented):
OK for now

PR #1982, opened 2024-08-11
Better migration plugin

RobinTail:
During my work on [allowed dependencies plugin](https://github.com/RobinTail/allowed-dependencies) I found a better way of distributing plugins.


RobinTail:
Should probably drop eslint 8 for that

RobinTail (commented):
ok

PR #1993, opened 2024-08-18
Ensure mime type on plain text responses

RobinTail:


RobinTail:
todo: add assertion to test

RobinTail (commented):
ok

PR #2035, opened 2024-09-12
Feat: Handling deprecation events by actual logger

RobinTail:
`express` uses `depd` for deprecating things.

This library emits a special event on process, and when there is a listener of that event, it can be handled other way than `stderr`.

https://www.npmjs.com/package/depd#processondeprecation-fn

```
2024-09-13T19:34:38.556Z warn: DeprecationError (express): res.location("back"): use res.location(req.get("Referrer") || "/") and refer to https://dub.sh/security-redirect for best practices [
  '    at file://.../example/config.ts:1:649',
  '    at Layer.handle [as handle_request] (.../node_modules/express/lib/router/layer.js:95:5)',
  '    at next (/.../node_modules/express/lib/router/route.js:149:13)',
  '    at Route.dispatch (/.../node_modules/express/lib/router/route.js:119:3)',
  '    at Layer.handle [as handle_request] (/.../node_modules/express/lib/router/layer.js:95:5)'
]
```

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -27,8 +27,6 @@ const makeCommonEntities = (config: CommonConfig) => {
     ? config.logger
     : new BuiltinLogger(config.logger);
   rootLogger.debug("Running", process.env.TSUP_BUILD || "from sources");
-  const onDeprecation = (err: Error) => rootLogger.warn(err.message, err);
-  process.on("deprecation", onDeprecation);
   const loggingMiddleware = createLoggingMiddleware({ rootLogger, config });
   const getChildLogger = makeChildLoggerExtractor(rootLogger);
   const commons = { getChildLogger, errorHandler };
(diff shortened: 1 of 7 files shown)


src/server.ts
```diff
@@ -27,6 +27,9 @@ const makeCommonEntities = (config: CommonConfig) => {
     ? config.logger
     : new BuiltinLogger(config.logger);
   rootLogger.debug("Running", process.env.TSUP_BUILD || "from sources");
+  process.on("deprecation", (error: Error) =>
```
RobinTail, line 30: this affects multiple unit tests because of the same process.
should be rather tested in system one.

RobinTail: done

src/server.ts
```diff
@@ -60,8 +66,10 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
     notFoundHandler,
     parserFailureHandler,
     loggingMiddleware,
+    onDeprecation,
   } = makeCommonEntities(config);
   const app = express().disable("x-powered-by").use(loggingMiddleware);
+  process.on("deprecation", onDeprecation);
```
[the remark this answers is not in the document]
RobinTail: addressed

RobinTail (commented):
ok

PR #2050, opened 2024-09-19
Exp: deconstructing config

RobinTail:


RobinTail:
not completely like it.
also there will be another change that will make it more readable #2086

PR #2064, opened 2024-09-30
Inverting definition of `AbstractLogger`

RobinTail:
Currently, `severity` is defined from `keyof AbstractLogger`.
I'd like an opposite arrangement.


RobinTail (commented):
nice

PR #2066, opened 2024-10-01
Fun: QR code for documentation

RobinTail:
This is fun :)

- no runtime dependencies: making the code during the build using its version number;
- using UTF-8 charset for `esbuild`.

image


RobinTail:
not planned, but cool

PR #2070, opened 2024-10-02
CI: coveralls: fixing version to 2.3.0

RobinTail:
failing to start from the tag `v2`


RobinTail:
does not fix the issue

PR #2099, opened 2024-10-15
Exp: HTTP/2

RobinTail:
Experiments regarding #2092 

- h2 is HTTP/2 over TLS (protocol negotiation via ALPN)
- no browsers supporting HTTP/2 without encryption
- Undici Agent option: `allowH2`

PR #2100, opened 2024-10-15
Return generic array of servers

RobinTail:
This should ease declaration and enable future improvements like #2099 

Making `createServer()` to return an object having `servers` property instead of `httpServer` and `httpsServer`


RobinTail (commented):
good

PR #2102, opened 2024-10-15
Fix: exposing `testMiddleware`

### LINKED ISSUE #2101: `testMiddleware` not exported
# Description

I tried to use `testMiddleware` which is mentioned in the documentation but it cannot be imported. Looking at `index.ts`, I found that only `testEndpoint` is currently exported.

See:
https://github.com/RobinTail/express-zod-api/blob/388cc941ef854e08a886e7832fc253bf822c4be6/src/index.ts#L32

## Expected

The `testMiddleware` function should be exported.

## Reproduction

Just try to import `testMiddleware` and it will fail.

image

## Context

- Node.js version: 20.14.0
- `express-zod-api` version: 20.14.2

RobinTail:
Fixes #2101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Version 20
 
+### v20.14.3
+
+- Fixed: missing export of `testMiddleware`:
+  - The feature introduced in v20.4.0 but was not available;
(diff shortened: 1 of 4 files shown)


RobinTail (commented):
ok

PR #2107, opened 2024-10-17
Add `module-sync` export condition

RobinTail:
Currently supported by Node 22.10 and 23.x
https://github.com/nodejs/node/releases/tag/v22.10.0

This should replace the `require` condition when all LTS versions support `require(ESM)`

22.12 enabled require(ESM) by default (without flag)


RobinTail:
- [ ] check this out before considering https://github.com/prettier/prettier/blob/main/CHANGELOG.md#352

RobinTail:
Doing #2814 instead

PR #2113, opened 2024-10-21
Explicit return types

RobinTail:
This should improve the DTS building performance

| Metric | Before | Header | Header |
|--------|--------|--------|--------|
| DTS Node 20.x | 4108ms | Cell | Cell |
| Cell | Cell | Cell | Cell |


RobinTail:
currently 4279ms, no visible effect yet

RobinTail:
4242ms, no benefit so far

RobinTail:
4287ms, it's not getting less

RobinTail:
4245ms, no benefit

PR #2115, opened 2024-10-22
Feat: support `Map` and `Set` serialization in JSON response

RobinTail:
Discussion #1500 

Making `Map` and `Set` serializable by using the Express application variable `json replacer`.
This, however, 2 times slower in runtime, so that it's going to be an opt-in feature, I believe.
Basically, ANY replacer makes `res.json()` twice slower.

```
   ✓ Experiment (4) 29107ms
     name                               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · no replacer              1,005,950.99  0.0009  0.8289  0.0010  0.0010  0.0012  0.0015  0.0153  ±0.44%   502976   fastest
   · empty replacer             553,796.75  0.0016  0.2556  0.0018  0.0019  0.0021  0.0024  0.0204  ±0.30%   276899
   · current implementation     489,327.54  0.0018  1.3553  0.0020  0.0020  0.0026  0.0027  0.0231  ±0.09%  4893276   slowest
   · featured implementation    492,527.85  0.0018  1.3013  0.0020  0.0020  0.0024  0.0026  0.0232  ±0.08%  4925279
```

PR #2116, opened 2024-10-24
Feat: deep check for unsupported JSON schema

RobinTail:
Implements #1500 
Instead of #2115 

```
2024-10-25T13:39:49.763Z warn: The final positive response schema of the endpoint contains an unsupported JSON payload type. {
  path: '/v1/user/list',
  method: 'get',
  reason: AssertionError [ERR_ASSERTION]: z.nan()
```


src/routing.ts
```diff
@@ -30,9 +34,30 @@ export const initRouting = ({
     routing,
     hasCors: !!config.cors,
     onEndpoint: (endpoint, path, method, siblingMethods) => {
+      const requestType = endpoint.getRequestType();
+      if (
+        requestType === "json" &&
+        hasJsonIncompatibleSchema(endpoint.getSchema("input"), false)
```
RobinTail, line 40: the hook can be called for the same `endpoint` reference multiple times when it's assigned to several routes, therefore it should be some optimization to reduce traversing

RobinTail: addressed

RobinTail (commented):
ok

RobinTail (commented):
brilliant

RobinTail:
✅ QA passed

PR #2117, opened 2024-10-25
Getting rid of serializers

RobinTail:
Serializers used to be involved for comparing schemas in order to handle possible circular references of ZodLazy.
It's time to make it better, using the power of `Map`.

### DIFF
--- a/src/documentation-helpers.ts
+++ b/src/documentation-helpers.ts
@@ -72,10 +72,12 @@ import { UploadSchema, ezUploadBrand } from "./upload-schema";
 
 export interface OpenAPIContext extends FlatObject {
   isResponse: boolean;
-  getRef: (schema: z.ZodTypeAny) => ReferenceObject | undefined;
   makeRef: (
     schema: z.ZodTypeAny,
-    depicted: SchemaObject | ReferenceObject,
+    image:
+      | SchemaObject
+      | ReferenceObject
+      | (() => SchemaObject | ReferenceObject),
     name?: string,
   ) => ReferenceObject;
   path: string;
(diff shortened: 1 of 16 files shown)


src/documentation.ts
```diff
@@ -81,19 +77,22 @@ export class Documentation extends OpenApiBuilder {
...
   protected lastOperationIdSuffixes = new Map<string, number>();
   protected responseVariants = keys(defaultStatusCodes);
+  protected references = new Map<z.ZodTypeAny, ReferenceObject>();
 
   protected makeRef(
-    name: string,
-    schema: SchemaObject | ReferenceObject,
+    schema: z.ZodTypeAny,
+    depicted: SchemaObject | ReferenceObject,
+    name = this.references.get(schema)?.$ref.split("/").pop() ??
```
RobinTail, line 85: should store just name, `make` should call `get`, and `get` should make an object with that name

RobinTail: fixed

src/documentation-helpers.ts
```diff
@@ -583,18 +580,10 @@ export const depictBranded: Depicter = (
...
-    (() => {
-      makeRef(hash, {}); // make empty ref first
-      return makeRef(hash, next(schema)); // update
-    })()
-  );
-};
+  lazy: z.ZodLazy<z.ZodTypeAny>,
+  { next, getRef, makeRef },
+): ReferenceObject =>
+  getRef(lazy) ?? (makeRef(lazy, {}) && makeRef(lazy, next(lazy.schema))); // create & update
```
RobinTail, line 586: `Map::get()` calls should be reduced:
1. getRef()
2. makeRef() x2

RobinTail: addressed

RobinTail (commented):
integration is ok

RobinTail (commented):
ok

PR #2129, opened 2024-10-29
Making `input` schema optional

RobinTail:
The most basic case — no inputs at all and I'd like to simplify the way it's addressed

### DIFF
--- a/src/endpoints-factory.ts
+++ b/src/endpoints-factory.ts
@@ -78,8 +78,8 @@ export class EndpointsFactory<
     ASCO extends string,
   >(
     subject:
-      | Middleware<AIN, OUT, AOUT, ASCO>
-      | ConstructorParameters<typeof Middleware<AIN, OUT, AOUT, ASCO>>[0],
+      | Middleware<OUT, AOUT, ASCO, AIN>
+      | ConstructorParameters<typeof Middleware<OUT, AOUT, ASCO, AIN>>[0],
   ) {
     return EndpointsFactory.#create<
       z.ZodIntersection<IN, AIN>,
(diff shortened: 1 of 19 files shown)


src/endpoints-factory.ts
```diff
@@ -78,8 +78,8 @@ export class EndpointsFactory<
     ASCO extends string,
   >(
     subject:
-      | Middleware<AIN, OUT, AOUT, ASCO>
-      | ConstructorParameters<typeof Middleware<AIN, OUT, AOUT, ASCO>>[0],
+      | Middleware<OUT, AOUT, ASCO, AIN>
+      | ConstructorParameters<typeof Middleware<OUT, AOUT, ASCO, AIN>>[0],
```
RobinTail, line 82: this case does not work properly without `input`

RobinTail: fixed in 3f0d817a

RobinTail (commented):
need more type tests

src/middleware.ts
```diff
@@ -28,10 +28,10 @@ export abstract class AbstractMiddleware {
 }
 
 export class Middleware<
-  IN extends IOSchema<"strip">,
   OPT extends FlatObject,
   OUT extends FlatObject,
   SCO extends string,
+  IN extends IOSchema<"strip"> = z.ZodObject<EmptyObject, "strip">,
```
RobinTail, line 34: this type and value should be worth extracting

RobinTail: does not work well for value, but type can be extracted

RobinTail: addressed in 7eb2d9d5

RobinTail:
✅ QA passed

RobinTail (commented):
ok

PR #2131, opened 2024-10-30
Merge intersected objects types in generated client

RobinTail:
This should prevent the "empty object type" in the generated client, which is not really a good type.
`Record<string, never>` can not be used in this case, so I'm trying to eliminate the root cause.


RobinTail (commented):
I like it

PR #2133, opened 2024-10-31
fix: check for output schema instead of variant

doiska:
Hello,

I'm using a custom ResultHandler based on top of the default one and having this issue:

```
 |  +2562047h   2024-10-31T03:46:42.614Z warn: The final positive response schema of the endpoint contains an unsupported JSON payload type. {   
                   │ |  +2562047h     path: '/locations/states',                                                                                                                                                                    
                   │ |  +2562047h     method: 'get',                                                                                                                                                                                
                   │ |  +2562047h     reason: TypeError: Cannot read properties of undefined (reading '_def')                                                                                                                       
                   │ |  +2562047h         at _e (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:4:3823)                                                                                              
                   │ |  +2562047h         at xt (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:4:4087)                                                                                              
                   │ |  +2562047h         at onEndpoint (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:5:10705)                                                                                     
                   │ |  +2562047h         at ee (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:5:9910)                                                                                              
                   │ |  +2562047h         at ee (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:5:10197)                                                                                             
                   │ |  +2562047h         at ee (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:5:10197)                                                                                             
                   │ |  +2562047h         at At (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:5:10358)                                                                                             
                   │ |  +2562047h         at yn (file:///home/eduardo/projects/docs/node_modules/express-zod-api/dist/index.js:21:437)                                                                                              
                   │ |  +2562047h         at <anonymous> (/home/eduardo/projects/docs/packages/backend/src/app/app.ts:34:29)                                                                                                        
                   │ |  +2562047h         at ModuleJob.run (node:internal/modules/esm/module_job:268:25)                                                                                                                            
                   │ |  +2562047h   }
```

I solved it by checking for "output" instead of "negative". I'm not familiar with this codebase, but since we're checking for input, I think we should be checking for output too. Its just a guess but solved my issue.

Result Handler:
```js
const resultHandler = new ResultHandler({
  positive: (output) =>
    z.object({
      status: z.literal("success"),
      data: output,
    }),
  negative: z.object({
    status: z.literal("error"),
    error: z.object({ message: z.string() }),
  }),
  handler: ({ error, input, output, request, response, logger }) => {
    if (!error) {
      response.status(200).json({ status: "success", data: output });
      return;
    }

    const statusCode = getStatusCodeFromError(error);
    response.status(statusCode).json({
      status: "error",
      error: { message: getMessageFromError(error) },
    });
  },
});
```


RobinTail:
Thank you for the feedback,

> @doiska 
> since we're checking for input, I think we should be checking for output too

`output` schema is the output of the endpoint, it's intermediate, and it's included/processed by the `positive` response schema of a ResultHandler that endpoint is using, making the final output schema for generators:

```ts
positive: (output) =>
    z.object({
      status: z.literal("success"),
      data: output,
    }),
```

So, what you changed here

```diff
        for (const variant of ["positive", "negative"] as const) {
          if (endpoint.getMimeTypes(variant).includes(contentTypes.json)) {
            try {
-              assertJsonCompatible(endpoint.getSchema(variant), "out");
+              assertJsonCompatible(endpoint.getSchema("output"), "out");
```

is unfortunately incorrect, because it's not enough for positive case and wrong for negative case (when there is an error instead of endpoint output), and because you're checking the same thing twice in the `for..of` loop.

Instead, the particular issue should be addressed that is taking place:

```
TypeError: Cannot read properties of undefined (reading '_def') 
```

Which means that somehow and for some reason some schema within the `positive` response structure is not `ZodType`, but `undefined`. 

I'd like also to check out your `new EndpointsFactory()` creation and what's assigned to the `output` of your Endpoint to figure that out.

RobinTail:
> node_modules/express-zod-api/dist/index.js:4:3823

This refers to the following statement in build:

```
let i=o<n?r[e._def[g]?.brand]||r[e._def.typeName]:void 0;
```

which seems to be `hasNestedSchema` statement in sources:
https://github.com/RobinTail/express-zod-api/blob/19e6af3dcf1853ea41e2334c0c2fda5be17b742a/src/deep-checks.ts#L76-L80

RobinTail:
> @doiska 
> Result Handler

I tried your Result Handler exactly as you posted above, and I'm not getting that error.
Its code looks correct. So that the issue is somewhere else related to the endpoint.

RobinTail:
I managed to achieve such error by doing this, @doiska : 

```ts
const doiskaResultHandler = new ResultHandler({
  positive: (output) => [{ schema: undefined }],
  // ...
});
```

RobinTail:
It's also possible to get such error when there is no `output` property inside `.build()` argument.

doiska:
The error only happens if I change the defaultResultHandler, even copying the entire code from the defaultResultHandler and creating new one doesn't work. And happens to all my endpoints, that were working before the v20.1 (iirc).
To give some context, im using lambda (arm64) and express v4.

This is my current result handler (working with my changes):

```js
const resultHandler = new ResultHandler({
  positive: (output) =>
    z.object({
      status: z.literal("success"),
      data: output,
    }),
  negative: z.object({
    status: z.literal("error"),
    error: z.object({ message: z.string() }),
  }),
  handler: ({ error, input, output, response, logger, request }) => {
    if (!error) {
      response.status(200).json({ status: "success", data: output });
      return;
    }

    logger.debug(`A handled error occurred at ${request.path}: `, {
      error,
      input,
      output,
    });

    if (!isHttpError(error)) {
      Sentry.captureException(error, {
        level: "error",
        tags: {
          method: request.method,
          path: request.path,
        },
      });
    }

    const statusCode = getStatusCodeFromError(error);
    response.status(statusCode).json({
      status: "error",
      error: { message: getMessageFromError(error) },
    });
  },
});
```

Even though is the wrong solution, its working for me. But i'm gonna remove the check for now.
Also, wanna move this to a issue instead? Since its the wrong solution.

image
image
image

Thanks for the fast response :)

doiska:
Forgot to mention, the EndpointFactory has 3 middlewares connected (auth stuff). But are the same middlewares that were connected to the defaultEndpointFactory.


And since you asked about the build:

image

GET /vendors/
```
import { protectedEndpointFactory } from "@/lib/endpoint-factory";
import { z } from "zod";
import { db } from "@/db";
import { dbVendors } from "@/db/schemas/vendors";
import { eq } from "drizzle-orm";

export const retrieveAllVendors = protectedEndpointFactory.build({
  method: "get",
  input: z.object({}),
  output: z.object({
    vendors: z.array(
      z.object({
        id: z.string(),
        name: z.string(),
        document: z.string(),
      })
    ),
  }),
  handler: async ({ options }) => {
    const vendors = await db
      .select({
        id: dbVendors.id,
        name: dbVendors.name,
        document: dbVendors.document,
      })
      .from(dbVendors)
      .where(eq(dbVendors.tenantId, options.tenantId));

    return {
      vendors,
    };
  },
});
```

RobinTail:
> even copying the entire code from the defaultResultHandler and creating new one doesn't work

that's very suspicios, @doiska 
Since I can not reproduce that issue, I have an idea: could you create a repo using `express-zod-api` as it is having such warning, so that I could clone it and debug?

RobinTail:
@doiska , meanwhile, I made a repo as well, using the code samples you provided:
https://github.com/RobinTail/issue2133

and there are no such warnings when I `yarn start` it

doiska:
For some reason, I cant replicate it anymore 🥲. I think was something related to lambda caching. I used to remove express-zod-api from bundling (dont remember why), but now that is bundling, the error its not happening anymore.
I can DM you somewhere the full code so you can check, but I think its working for now.

Thanks

RobinTail:
Alright, @doiska , I'm glad that we figured that out.

PR #2135, opened 2024-10-31
Switch coverage provider to v8

RobinTail:
- Reducing dependencies incl. removal of outdated ones,
- More strict coverage calculation


RobinTail (commented):
I like the overall deps tree improvement

RobinTail (commented):
ok

PR #2136, opened 2024-11-01
Exp: Non-disclosure of internal errors

RobinTail:
1. ~~I wanna use `cause` for wrapping errors~~
  a. moved to #2137 
3. To improve security, internal errors should not be disclosed in response, but logged completely
4. I wanna give to those errors more meaningful details
5. This should probably be configurable


src/last-resort.ts
RobinTail, line 1: should also use the config option

RobinTail: done

RobinTail:
I think I should split this.
Too many large deviation.

RobinTail:
need to start this over

PR #2137, opened 2024-11-02
Replacing `originalError` with `cause` on validation errors

RobinTail:
Cherry-picked from #2136 

`cause` is ES2022 feature, its support introduced in Node.js 16.9 

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#browser_compatibility
- https://nodejs.org/api/errors.html#errorcause

### DIFF
--- a/tests/unit/errors.spec.ts
+++ b/tests/unit/errors.spec.ts
@@ -67,7 +67,8 @@ describe("Errors", () => {
       expect(error.name).toBe("OutputValidationError");
     });
 
-    test("should have .originalError property matching the one used for constructing", () => {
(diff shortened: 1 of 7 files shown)


tests/unit/errors.spec.ts
```diff
@@ -86,6 +87,7 @@ describe("Errors", () => {
     });
 
     test("should have .originalError property matching the one used for constructing", () => {
```
RobinTail, line 89:
```suggestion
    test("should have .cause property matching the one used for constructing", () => {
```

RobinTail (commented):
ok

PR #2139, opened 2024-11-02
Drop `originalError` property on validation errors

RobinTail:
Due to #2137 deprecated

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -41,14 +41,6 @@ export class OutputValidationError extends IOSchemaError {
   constructor(public override readonly cause: z.ZodError) {
     super(getMessageFromError(cause), { cause });
   }
-
-  /**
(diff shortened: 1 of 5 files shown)


src/errors.ts
```diff
@@ -41,14 +41,6 @@ export class OutputValidationError extends IOSchemaError {
   constructor(public override readonly cause: z.ZodError) {
     super(getMessageFromError(cause), { cause });
   }
-
-  /**
-   * @deprecated use the cause property instead
-   * @todo remove in v21
-   * */
-  public get originalError() {
```
RobinTail, line 49: add migration script

RobinTail: to be sure we need to enforce type checking, but it complicates the whole thing significantly

RobinTail (commented):
ok

PR #2141, opened 2024-11-02
Changing the `cause` on `ResultHandlerError`

RobinTail:
Adding another prop to hold the error that was processed by the ResultHandler when it failed.
Partially cherry-picked from #2136 , but this one has no regression


RobinTail (commented):
ready

PR #2142, opened 2024-11-02
Better error logging

RobinTail:
Complete error objects in logs instead of only `message`.

Cherry-picked from #2136


RobinTail (commented):
ok

PR #2144, opened 2024-11-03
Generalizing server-side error messages in production mode by default

RobinTail:
- Instead of #2136 
- Does not expose exact error messages to response when:
  - `NODE_ENV === "production"` and `error.expose === false` (all status codes in `5XX` block by default)
- This also deprecates `getStatusCodeFromError()` in favour of `ensureHttpError().statusCode`
- Switching to operating `HttpError` in `defaultResultHandler`
- Showing the `NODE_ENV` on startup
- https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production
- [x] also need to update Readme / Error handling section
- [x] AND lastResortHandler should follow this concept as well
- [x] Need MORE tests for last resort handler

PR #2149, opened 2024-11-04
Exp(21): Operate `HttpError` in Result Handlers

RobinTail:
Due to #2144 

- It might be even better to `ensureHttpError()` internally, so that won't be required to call it in every result handler.
- [x] however, some custom handling should still be possible to customize, presumably in config
- [ ] migration in that case should be changed

### DIFF
--- a/src/config-type.ts
+++ b/src/config-type.ts
@@ -49,6 +49,12 @@ export interface CommonConfig<TAG extends string = string> {
    */
   errorHandler?: AbstractResultHandler;
   /**
+   * @desc Interprets an error into a status code for HttpError handled by a ResultHandler
+   * @default error instanceof InputValidationError ? 400 : 500
+   * @todo naming
+   * */
+  getStatusCode?: (error: Error) => number;
+  /**
    * @desc Built-in logger configuration or an instance of any compatible logger.
    * @example { level: "debug", color: true }
    * */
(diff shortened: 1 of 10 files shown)


RobinTail:
MAYBE!
Just maybe.
It may be that we can just remove custom errors and always use HttpError, but allow to configure it for certain cases.
Just a though

src/config-type.ts
```diff
@@ -48,6 +48,12 @@ export interface CommonConfig<TAG extends string = string> {
    * @see defaultResultHandler
    */
   errorHandler?: AbstractResultHandler;
+  /**
+   * @desc Interprets an error into a status code for HttpError handled by a ResultHandler
+   * @default error instanceof InputValidationError ? 400 : 500
+   * @todo naming
+   * */
+  getStatusCode?: (error: Error) => number;
```
RobinTail, line 56: MAYBE!
Just maybe.
It may be that we can just remove custom errors and always use HttpError, but allow to configure it for certain cases.
Just a though

RobinTail: The interfaces might get easier, no extra processing and wrapping.
Think about it.

RobinTail (commented):
despite the neatness for default result handler, I'm not going to proceed with this, because it seems to be too opinionated and I don't wanna take from developers or complicate the decision on how to treat certain events considered by the internal flow as Input/OutputValidationError.
Let's keep the status code determination in ResultHandler only

PR #2154, opened 2024-11-06
Feat: ability to catch errors in `testMiddleware`

### LINKED ISSUE #2153: Catch errors in testMiddleware function
It's common for middleware to throw exceptions when validating auth tokens or other request information. When this happens, the mocks available within `testMiddleware` won't be returned and can't be used to confirm behavior. This makes it difficult to test behavior of anything that happened up to the point of failure. For example:

```javascript
import { z } from "zod";
import { Middleware, testMiddleware } from "express-zod-api";

const middleware = new Middleware({
  input: z.object({ test: z.string() }),
  handler: async ({ options, input: { test }, logger }) => {
     logger.info("logging something");
     throw new Error("something went wrong");
  },
});

// EXCEPTION THROWN HERE
const { output, responseMock, loggerMock } = await testMiddleware({
  middleware,
  requestProps: { method: "POST"},
  options: { prev: "accumulated" }, 
});
expect(loggerMock._getLogs().info).toHaveLength(1);  
```

To implement this we could provide an extra field to the function to indicate the intention of getting an error response back.  Something like:
```javascript
const { output, responseMock, loggerMock, error } = await testMiddleware({ 
  middleware,
  requestProps: { method: "POST"},
  options: { prev: "accumulated" }, 
  catchErrors: true, // when true will catch errors thrown by middleware and return them in the response
});
expect(loggerMock._getLogs().info).toHaveLength(1);
```

RobinTail:
Closes #2153


williamgcampbell (approved):
Thank you!

PR #2168, opened 2024-11-14
Traverse without recursion

RobinTail:
Avoiding recursion should decrease memory consumption by replacing the utilization of call stack to a stack of entities to process

https://www.nearform.com/insights/tracking-memory-allocation-node-js/

```
   ✓ Experiment for routing walker (2) 1247ms
     name             hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · current    3,691.20  0.2053  2.7257  0.2709  0.2885  0.3817  0.4519  2.1826  ±1.49%     1846
   · featured  50,253.26  0.0123  5.8032  0.0199  0.0165  0.0423  0.1116  2.7895  ±7.10%    25127   fastest
```

Turned out that the performance is mostly affected by `assert(..., new Error())` pattern, partially introduced by https://github.com/RobinTail/express-zod-api/pull/1371

PR #1371, opened 2023-12-10
Ref: using `node:assert/strict` instead of `if + throw`

RobinTail:
Reducing branch complexity that way.
⚠️ Requires to get rid of Jest: #1369

PR #2169, opened 2024-11-15
Revert: `throw` instead of `assert`

RobinTail:
partially reverts https://github.com/RobinTail/express-zod-api/pull/1371 for performance reasons found in #2168


RobinTail (commented):
good

PR #2172, opened 2024-11-15
Addressing performance findings

RobinTail:
I wanna put all my recent performance findings into ESLint rules, so that I could address them globally later

- Object.assign turned out to be significantly improved in Node 22, but it's still faster for extending existing aggregator in a reducer. For other cases spread is ok.


RobinTail:
should be enough so far

RobinTail (commented):
ok

PR #2174, opened 2024-11-16
Generalising server type

RobinTail:
`http.Server | https.Server` —> `net.Server`, common ancestor


RobinTail (commented):
ok

RobinTail:
reverted it later, because it turned out that it can complicate the consumption

PR #2175, opened 2024-11-16
Universal properties for arrays

RobinTail:
New attempt for #1536 
Based on the discussion https://github.com/RobinTail/express-zod-api/discussions/1193#discussioncomment-8014231

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -42,7 +42,7 @@ export type { FlatObject } from "./common-helpers";
 export type { Method } from "./method";
 export type { IOSchema } from "./io-schema";
 export type { CommonConfig, AppConfig, ServerConfig } from "./config-type";
-export type { ApiResponse } from "./api-response";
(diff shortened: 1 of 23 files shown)


src/index.ts
```diff
@@ -42,7 +42,7 @@ export type { FlatObject } from "./common-helpers";
 export type { Method } from "./method";
 export type { IOSchema } from "./io-schema";
 export type { CommonConfig, AppConfig, ServerConfig } from "./config-type";
-export type { ApiResponse } from "./api-response";
+export type { ApiResponse, NormalizedResponse } from "./api-response";
```
RobinTail, line 45: maybe should still express NormalizedResponse via ApiResponse to remove it

RobinTail: fixed

RobinTail:
The experiment succeeded.
Evaluating the idea of including it into v21...

RobinTail (commented):
ok, I think it's better to do it now than unclear when

PR #2128, opened 2024-10-29
Making `method` and `methods` optional

RobinTail:
- Default should be `get`;
- I'm also thinking to make it optional for `DependsOnMethod` case
  - then the default would be the property of its constructor the endpoint assigned to
  - for that, `getMethods()` of Endpoint would return empty (which technically it can) or `undefined`, and the decision making on method would be delegated to Routing Walker.
  - this, however, deviates from the initial concept and affect such features as `testEndpoint` for CORS enabled endpoints, so that it could not respond with its allowed methods without routing, OR decetralizing the method decision making.
  - the workaround might be moving CORS headers handling into `initRouting`

PR #2176, opened 2024-11-16
Removing redundant methods array in tests.

RobinTail:
Due to #2128 , did not found those lines initially, but found during work in #2175


RobinTail (commented):
good

PR #2179, opened 2024-11-16
Client: JSON parser in case of path params

### LINKED ISSUE #2177: Error in frontend client usage exemple
# Description

The usage example in the generated client tries to find out if the response should be parsed as JSON or text using this line of code:
 ```
const parser = `${method} ${path}` in jsonEndpoints ? "json" : "text";
 ```

But `path` in this context is not a `MethodPath` but the path where the parameters have already been replaced by their value. Because of that, the "json" parser will never be selected

## Expected

It's not clear to me what the cleanest fix would be. I suppose `Implementation` could accept a 4th param that would be the original `MethodPath` ?

## Reproduction

Try to use the generated client with a JSON endpoint that has a path parameter

## Context

- Node.js version: 22
- `express-zod-api` version: 20.21.1

## Additional context

Love this lib! Really incredible work!

RobinTail:
Fixes #2177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## Version 20
 
+### v20.21.2
+
+- Fixed the example implementation in the generated client for endpoints using path params:
+  - The choice of parser was made based on the exported `const jsonEndpoints` indexed by `path`;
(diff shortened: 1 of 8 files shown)


RobinTail (commented):
ok

PR #2185, opened 2024-11-17
Ease `DocumentationError` constructor

RobinTail:
It's publicly exposed, therefore I consider it breaking.
I wanna make the usage of that error easier.


RobinTail (commented):
ok

PR #2191, opened 2024-11-18
Feat: Empty response support

RobinTail:
The following HTTP status codes imply no content in response:
- 204
- redirects, such as 302

First one usually used for REST deletions. 

There should be an easy way to describe and such things. Basically, there is no MIME type for those cases, because there is no body.

Current obstacles:
- `ResultHandler::positive::mimeTypes` is currently non-empty array, either `mimeType` is a string
- `z.never()` is probably the best candidate for the `schema` but it's unsupported
  - ~~however, `not: {}` is used for describing fixed tuples, which the best depiction~~

Consideration:
- ~~if `never` type should be used in the client, because of no body~~
- check the presence of content type in example impl.
- or maybe it should be `undefined`, because `never` means that `.provide()` does not return at all, such as `throw` case
- or maybe response type should not be processed by ZTS at all, similar to the desired OpenAPI no-MIME-type case
- probably it should be the feature of v21 having less variation on MIME type declaration

### DIFF
--- a/example/endpoints/delete-user.ts
+++ b/example/endpoints/delete-user.ts
@@ -1,0 +1,22 @@
+import createHttpError from "http-errors";
+import assert from "node:assert/strict";
+import { z } from "zod";
+import { noContentFactory } from "../factories";
+
+/** @desc The endpoint demonstrates no content response */
+export const deleteUserEndpoint = noContentFactory.build({
+  method: "delete",
+  tag: "users",
+  input: z.object({
+    id: z
+      .string()
+      .regex(/\d+/)
+      .transform((id) => parseInt(id, 10))
+      .describe("numeric string"),
+  }),
+  output: z.object({}),
+  handler: async ({ input: { id } }) => {
+    assert(id <= 100, createHttpError(404, "User not found"));
+    return {};
+  },
+});
--- a/example/endpoints/delete-user.ts
+++ b/example/endpoints/delete-user.ts
@@ -3,7 +3,7 @@ import assert from "node:assert/strict";
 import { z } from "zod";
 import { noContentFactory } from "../factories";
 
-/** @desc The endpoint demonstrates no content response */
+/** @desc The endpoint demonstrates no content response established by its factory */
 export const deleteUserEndpoint = noContentFactory.build({
   method: "delete",
   tag: "users",
--- a/example/example.client.ts
+++ b/example/example.client.ts
@@ -27,6 +27,35 @@ type GetV1UserRetrieveResponse =
       };
     };
 
+type DeleteV1UserIdRemoveInput = {
+  /** numeric string */
+  id: string;
+};
+
+type DeleteV1UserIdRemoveResponse = undefined;
+
+type PatchV1UserIdInput = {
+  key: string;
+  id: string;
+  name: string;
+  birthday: string;
+};
+
+type PatchV1UserIdResponse =
+  | {
+      status: "success";
+      data: {
+        name: string;
+        createdAt: string;
+      };
+    }
+  | {
+      status: "error";
+      error: {
+        message: string;
+      };
+    };
+
 type PostV1UserCreateInput = {
   name: string;
 };
@@ -107,45 +136,16 @@ type PostV1AvatarRawResponse =
       };
     };
 
-type DeleteV1UserIdRemoveInput = {
-  /** numeric string */
-  id: string;
-};
-
-type DeleteV1UserIdRemoveResponse = undefined;
-
-type PatchV1UserIdInput = {
-  key: string;
-  id: string;
-  name: string;
-  birthday: string;
-};
-
-type PatchV1UserIdResponse =
-  | {
-      status: "success";
-      data: {
-        name: string;
-        createdAt: string;
-      };
-    }
-  | {
-      status: "error";
-      error: {
-        message: string;
-      };
-    };
-
 export type Path =
   | "/v1/user/retrieve"
+  | "/v1/user/:id/remove"
+  | "/v1/user/:id"
   | "/v1/user/create"
   | "/v1/user/list"
   | "/v1/avatar/send"
   | "/v1/avatar/stream"
   | "/v1/avatar/upload"
-  | "/v1/avatar/raw"
-  | "/v1/user/:id/remove"
-  | "/v1/user/:id";
+  | "/v1/avatar/raw";
 
 export type Method = "get" | "post" | "put" | "delete" | "patch";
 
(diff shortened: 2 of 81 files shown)


example/example.documentation.yaml
```diff
@@ -85,6 +85,33 @@ paths:
...
+            type: string
+            pattern: \d+
+            description: numeric string
+      responses:
+        "204":
+          description: DELETE /v1/user/:id/remove Positive response
+          content:
+            text/plain:
+              schema:
+                not: {}
```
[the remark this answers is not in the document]
RobinTail: addressed

example/example.client.ts
```diff
@@ -27,6 +27,13 @@ type GetV1UserRetrieveResponse =
       };
     };
 
+type DeleteV1UserIdRemoveInput = {
+  /** numeric string */
+  id: string;
+};
+
+type DeleteV1UserIdRemoveResponse = never;
```
RobinTail, line 35: should be `undefined` because it's not an error

src/documentation-helpers.ts
```diff
@@ -413,7 +413,7 @@ export const depictTuple: Depicter = (
 ) => ({
   type: "array",
   prefixItems: items.map(next),
-  // does not appear to support items:false, so not:{} is a recommended alias
+  // does not appear to support items:false, so not:{} is a recommended alias	  items: rest === null ? { not: {} } : next(rest), // does not support items:false
   items: rest === null ? { not: {} } : next(rest),
```
RobinTail, line 417:
```suggestion
  // does not appear to support items:false, so not:{} is a recommended alias
  items: rest === null ? { not: {} } : next(rest),
```

src/documentation-helpers.ts
```diff
@@ -802,22 +802,27 @@ export const depictResponse = ({
...
     schema:
       composition === "components"
         ? makeRef(schema, depictedSchema, makeCleanId(description))
         : depictedSchema,
     examples: depictExamples(schema, true),
   };
-  return { description, content: fromPairs(xprod(mimeTypes, [media])) };
+  return {
+    description,
+    content: media && fromPairs(xprod(mimeTypes, [media])),
```
RobinTail, line 824: probably I could simplify that using `if` conditions

example/example.client.ts
RobinTail: reverted to any so far...

example/example.client.ts
```diff
@@ -27,6 +27,13 @@ type GetV1UserRetrieveResponse =
       };
     };
 
+type DeleteV1UserIdRemoveInput = {
+  /** numeric string */
+  id: string;
+};
+
+type DeleteV1UserIdRemoveResponse = any;
```
RobinTail, line 35: should be `undefined` because `.provide()` does not throw

RobinTail: fixed

src/api-response.ts
```diff
@@ -11,8 +11,11 @@ export interface ApiResponse<S extends z.ZodTypeAny> {
   schema: S;
   /** @default 200 for a positive and 400 for a negative response */
   statusCode?: number | [number, ...number[]];
-  /** @default "application/json" */
-  mimeType?: string | [string, ...string[]];
+  /**
+   * @example [] for no content, such as 204 and 302
+   * @default "application/json"
+   * */
+  mimeType?: string | string[];
```
RobinTail, line 18: what would it take to make it `null` instead of empty array?
I think it could be more intuitively understandable

src/integration.ts
```diff
@@ -173,7 +173,9 @@ export class Integration {
         const positiveResponseId = splitResponse
           ? makeCleanId(method, path, "positive.response")
           : undefined;
-        const positiveSchema = endpoint.getSchema("positive");
+        const positiveSchema = endpoint.getMimeTypes("positive").length
+          ? endpoint.getSchema("positive")
+          : z.undefined();
```
RobinTail, line 178: extract/configure?

example/example.client.ts
```diff
@@ -129,60 +107,93 @@ type PostV1AvatarRawResponse =
...
+  | "/v1/user/:id/remove"
+  | "/v1/user/:id";
 
 export type Method = "get" | "post" | "put" | "delete" | "patch";
 
 export type MethodPath = `${Method} ${Path}`;
 
 export interface Input extends Record<MethodPath, any> {
   "get /v1/user/retrieve": GetV1UserRetrieveInput;
-  "patch /v1/user/:id": PatchV1UserIdInput;
```
RobinTail, line 148: can we preserve the position in case `.nest()` used?

RobinTail: addressed in #2192

example/endpoints/delete-user.ts
RobinTail, line 1: needs a system test for that endpoint

example/example.client.ts
RobinTail, line 1: example implementation needs `.bodyUsed` check

src/documentation-helpers.ts
RobinTail: addressed

src/integration.ts
RobinTail: extracted

example/example.client.ts
RobinTail: done

example/endpoints/delete-user.ts
RobinTail: done

src/api-response.ts
RobinTail: addressed in #2193

RobinTail: fixed in 77731539

src/api-response.ts
```diff
@@ -21,5 +24,8 @@ export interface ApiResponse<S extends z.ZodTypeAny> {
 
 export type NormalizedResponse = Pick<ApiResponse<z.ZodTypeAny>, "schema"> & {
   statusCodes: Extract<ApiResponse<z.ZodTypeAny>["statusCode"], Array<unknown>>;
-  mimeTypes: Extract<ApiResponse<z.ZodTypeAny>["mimeType"], Array<unknown>>;
+  mimeTypes: Extract<
+    ApiResponse<z.ZodTypeAny>["mimeType"],
+    Array<unknown> | null
+  >;
```
RobinTail, line 30: constraints are awesome, but these look overcomplicated

src/result-helpers.ts
```diff
@@ -43,11 +47,21 @@ export const normalize = <A extends unknown[]>(
...
           ? [mimeType]
-          : mimeType || features.mimeTypes,
+          : mimeType === undefined
+            ? features.mimeTypes
+            : mimeType,
     }),
   );
 };
 
+export const emptyResponse = (
```
RobinTail, line 57: perhaps we don't need it, just make an article in Readme instead?

RobinTail: done

src/api-response.ts
RobinTail: addressed in #2194

RobinTail (commented):
good, it's very straightforward now

RobinTail:
🚀 v21.1.0
✅ QA passed

RobinTail (commented):
ready

PR #2193, opened 2024-11-19
Drop `Endpoint::getMimeTypes()`

RobinTail:
needed for #2191 as a better way to handle responses array


src/result-helpers.ts
```diff
@@ -24,7 +24,15 @@ export const normalize = <A extends unknown[]>(
...
     return normalize(subject(...features.arguments), features);
-  if (subject instanceof z.ZodType) return [{ ...features, schema: subject }];
+  if (subject instanceof z.ZodType) {
+    return [
+      {
+        schema: subject,
+        mimeTypes: features.mimeTypes,
+        statusCodes: features.statusCodes,
+      },
+    ];
```
RobinTail, line 34: can it be shorter?

RobinTail (commented):
awesome

src/result-helpers.ts
RobinTail: done

PR #2195, opened 2024-11-19
Vitest snapshot serializers instead of traversing schema in tests

RobinTail:
I could now use the native vitest object traverse for that

https://vitest.dev/guide/snapshot.html#custom-serializer

No performance degradation detected


RobinTail (commented):
I consider it a success

PR #2196, opened 2024-11-19
feat(v21): Async verification of endpoints

RobinTail:
- it should look better
- verification is not a deal breaker for the startup, so that it can be postponed
- https://nodejs.org/en/learn/asynchronous-work/understanding-setimmediate
- this way it won't run until the process event loop is free


RobinTail:
should rather disable it in production mode

Issue #2182, opened 2024-11-17
Generated frontend client allows calling endpoint that do not exist


HenriJ:
# Description

In the current implementation `MethodPath` is the combination of all possible methods with all possible paths.

What it means is that I am able to call
`client.provide("delete", "/v1/user/retrieve", { id: "10" });`

even if there is no delete method linked with this path.

## Expected

I should get a typescript error when calling a path that does not provide the requested method.

As a workaround, I implemented my own client that instead of taking a method and a path separately directly takes the full endpoint "get /v1/user/retrieve" (I'm using `keyof type endpointTags` to get the list of available endpoints which is not perfect though).

## Reproduction

Try to call an endpoint with a method that does not exists.

## Context

- Node.js version: 22
  - `express-zod-api` version: 20.21.0

RobinTail:
> I am able to call

well, this is a conceptual problem, because technically you do actually can, but the request might be handled by "not found (404)" handler which is a part of the API. So, what I mean is that the ability to request an unlisted path/method is not a bug, but a property of the API. This behaviour can be useful when there is routing besides the one established by the framework itself, which is possible when using `attachRouting()` to an existing app, or by using `beforeRouting` feature in config. Thus, when you type `client.provide()` you may notice that the return type is `any` instead of something particular.

However, I understand the confusion.

> my own client ... takes ...  "get /v1/user/retrieve"

this is similar to the interface of [OctoKit Request](https://github.com/octokit/octokit.js?tab=readme-ov-file#octokitrequest).
I deliberately wanted to separate them.

Here is what I can do:

1. Stop extending `Response` from combinations and treat it as limited dictionary with literal keys

```diff
- interface Response extends Record<MethodPath, any>
+ interface Response
```

2. Change the `Provider` return type

```diff
- Promise<Response[`${M} ${P}`]>
+ Promise<`${M} ${P}` extends keyof Response ? Response[`${M} ${P}`] : never>
```

And probably this `never` should be configurable or at least opt-in feature.

RobinTail:
I check how OctoKit behaves in a similar scenario, @HenriJ , which I'd consider an etalon of REST API client, and it turned out that it does not prohibit unknown methods and undeclared routes. It acts exactly the same way as I implemented here — it returns `any`

![Screenshot 2024-11-17 at 12 24 34](https://github.com/user-attachments/assets/d3fcbc67-db78-40ba-bd64-fde25b2d56bb)

[...]

PR #2200, opened 2024-11-20
Client: handling excessive routes

RobinTail:
Address #2182 and #2183

### DIFF
--- a/example/example.client.ts
+++ b/example/example.client.ts
@@ -202,7 +202,9 @@ export type Provider = <M extends Method, P extends Path>(
   params: `${M} ${P}` extends keyof Input
     ? Input[`${M} ${P}`]
     : Record<string, any>,
-) => Promise<`${M} ${P}` extends keyof Response ? Response[`${M} ${P}`] : any>;
+) => Promise<
+  `${M} ${P}` extends keyof Response ? Response[`${M} ${P}`] : never
+>;
 
 export type Implementation = (
   method: Method,
--- a/example/example.client.ts
+++ b/example/example.client.ts
@@ -151,7 +151,7 @@ export type Method = "get" | "post" | "put" | "delete" | "patch";
 
 export type MethodPath = `${Method} ${Path}`;
 
-export interface Input extends Record<MethodPath, any> {
+export interface Input {
   "get /v1/user/retrieve": GetV1UserRetrieveInput;
   "delete /v1/user/:id/remove": DeleteV1UserIdRemoveInput;
   "patch /v1/user/:id": PatchV1UserIdInput;
@@ -163,7 +163,7 @@ export interface Input extends Record<MethodPath, any> {
   "post /v1/avatar/raw": PostV1AvatarRawInput;
 }
 
-export interface Response extends Record<MethodPath, any> {
+export interface Response {
   "get /v1/user/retrieve": GetV1UserRetrieveResponse;
   "delete /v1/user/:id/remove": DeleteV1UserIdRemoveResponse;
   "patch /v1/user/:id": PatchV1UserIdResponse;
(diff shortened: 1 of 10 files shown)


example/example.client.ts
```diff
@@ -199,8 +199,10 @@ export const endpointTags = {
 export type Provider = <M extends Method, P extends Path>(
   method: M,
   path: P,
-  params: Input[`${M} ${P}`],
-) => Promise<Response[`${M} ${P}`]>;
+  params: `${M} ${P}` extends keyof Input
+    ? Input[`${M} ${P}`]
+    : Record<string, any>,
+) => `${M} ${P}` extends keyof Response ? Promise<Response[`${M} ${P}`]> : any;
```
RobinTail, line 205: this should be Promise of condition instead

RobinTail: 02f48a0d

RobinTail:
I've got another idea...

RobinTail:
I'm still struggling to figure out what exactly it should be.
The original statement was that it should be Typescript error.
Meanwhile, technically it's still possible to make unlisted requests, and the response could even be typed from `errorHandler`.

I need to decide on the goal:
- should it be prevention of manual input mistakes
- or should it be correctness

RobinTail:
The overload approach

```ts
export class ExpressZodAPIClient {
  constructor(protected readonly implementation: Implementation) {}
  public provide<M extends Method, P extends Path>(
    method: M,
    path: P,
    params: `${M} ${P}` extends keyof Input
      ? Input[`${M} ${P}`]
      : Record<string, any>,
  ): Promise<
    `${M} ${P}` extends keyof Response ? Response[`${M} ${P}`] : unknown
  >;
  public provide<K extends keyof Input>(request: K, params: Input[K]): Promise<K>;
  public provide(...args: [string, string, Record<string, any>] | [string, Record<string, any>]) {
    const [method, path, params] = (args.length === 3 ? args : [...args[0].split(" "), args[1]]) as [Method, Path, Record<string, any>];
    return this.implementation(
      method,
      Object.keys(params).reduce(
        (acc, key) =>
          acc.replace(`:${key}`, params[key]),
        path,
      ),
      Object.keys(params).reduce(
        (acc, key) => Object.assign( acc, !path.includes(`:${key}`) && { [key]: params[key] } ),
        {},
      ),
    );
  }
}
```

RobinTail:
should also:
- deprecate `MethodPath`
- redefine and deprecate `Provider`

example/example.client.ts
```diff
@@ -210,21 +204,48 @@ export type Implementation = (
...
+            !path.includes(`:${key}`) && { [key]: params[key] },
+          ),
         {},
       ),
     );
+  }
 }
 
+// @todo add:
+// export type Provider = ExpressZodAPIClient["provide"];
```
RobinTail, line 247: todo

RobinTail: done

src/integration.ts
```diff
@@ -382,51 +364,187 @@ export class Integration {
       [f.createTemplateSpan(this.ids.keyParameter, emptyTail)],
     );
 
-    // Object.keys(params).reduce((acc, key) => acc.replace(___, params[key]), path)
+    // Object.keys(params).reduce((acc, key) => acc.replace(___, params[key as keyof typeof params]), path)
```
RobinTail, line 367:
```suggestion
    // Object.keys(params).reduce((acc, key) => acc.replace(___, params[key]), path)
```

example/example.client.ts
```diff
@@ -210,21 +204,49 @@ export type Implementation = (
...
+    request: K,
+    params: Input[K],
+  ): Promise<Response[K]>;
+  public provide(
+    ...args:
+      | [string, string, Record<string, any>]
+      | [string, Record<string, any>]
+  ) {
+    const [method, path, params] = (
+      args.length === 2 ? [...args[0].split(" "), args[1]] : args
```
RobinTail, line 227: I'm concerned about this `.split(" ")`.
I suppose it's possible for a path to contain spaces

RobinTail: addressed in 68e0ff04

RobinTail (commented):
ok, now need:
- beta
- changelog
- check readme for needed update

RobinTail:
🚀 21.2.0-beta.1

RobinTail:
✅ QA passed by https://github.com/RobinTail/express-zod-api/discussions/2183#discussioncomment-11427304

PR #2202, opened 2024-11-22
Requesting OpenAPI validation

RobinTail:
Caused by failures happening to the Docker hosted validator in #2201 

Resources
- https://github.com/badges/shields/blob/master/services/swagger/swagger.service.js
- https://validator.swagger.io/#/


RobinTail:
it should not run for every Node.js in matrix

PR #2180, opened 2024-11-16
Add comparison with tRPC to README

HenriJ:
Following discussion in https://github.com/RobinTail/express-zod-api/discussions/2178

PR #2203, opened 2024-11-23
Comparison with tRPC

RobinTail:
Following discussion in
https://github.com/RobinTail/express-zod-api/discussions/2178

Copied from #2180


RobinTail:
I had a lot of considerations in this regard and came to conclusion that I don't wanna put into README a comparison to anything that
- I have no expertise;
- Can be changed;
- Is not a reason for creating this framework.

What I would be interested to include into documentation instead is a migration guide "from *** to express-zod-api".

CC @HenriJ

PR #2225, opened 2024-12-01
Always split response type

RobinTail:
Partially based on discussion https://github.com/RobinTail/express-zod-api/discussions/2219
Needed for #2226 
But generally should simplify the flow and enable further improvements of the Integration class

PR #2226, opened 2024-12-01
Generating `EncodedResponse` interface by `Integration`

RobinTail:
Should go after #2225 

Based on discussion https://github.com/RobinTail/express-zod-api/discussions/2219


RobinTail:
I'm not sure it would allow making status code operating client.
It would require a single 3-dimensional dictionary-like interface.

example/example.client.ts
```diff
@@ -20,13 +22,31 @@ type GetV1UserRetrievePositiveResponse = {
   };
 };
 
-type GetV1UserRetrieveNegativeResponse = {
+interface GetV1UserRetrievePositiveResponseVariants {
+  200: GetV1UserRetrievePositiveVariant1;
+}
+
+type GetV1UserRetrievePositiveResponse =
+  SomeOf<GetV1UserRetrievePositiveResponseVariants>;
```
RobinTail, line 30: maybe I can reduce the number of entities by allowing public interfaces to hold type expressions, not only type references

RobinTail: addressed

RobinTail (commented):
OK in general

example/example.client.ts
```diff
@@ -3,12 +3,14 @@ type Type1 = {
   features: Type1;
 }[];
 
+export type SomeOf<T> = T[keyof T];
```
RobinTail, line 6: why public?

RobinTail (commented):
✅ QA passed

PR #2234, opened 2024-12-03
Use entity references instead of id holding constants

RobinTail:
This should ease further improvements like #2226, reduce the amount of declarations and improve overall readability


RobinTail (commented):
nice

PR #2235, opened 2024-12-03
Deduplicating type making helper

RobinTail:
There's been two methods for making a type: public and internal, with and without commenting possibility.
I'm combining them finally.


src/integration-helpers.ts
```diff
@@ -1,5 +1,6 @@
 import ts from "typescript";
 import { Method } from "./method";
+import { addJsDocComment } from "./zts-helpers";
```
RobinTail, line 3: this is fine, because the utility is involved into low level ZTS production of commented props

RobinTail (commented):
ok

PR #2238, opened 2024-12-04
SSE

RobinTail:
This should be an alternative to `zod-sockets` and enable subscriptions without websockets, without socket.io and so on. Also related to #991 

### Resources

- Standard: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format
- Article: https://web.dev/articles/eventsource-basics
- stack: https://stackoverflow.com/questions/34657222/how-to-use-server-sent-events-in-express-js
- random tutorial: https://masteringjs.io/tutorials/express/server-sent-events

### Findings

- it turned out it's required to do `flush()` after each message: https://stackoverflow.com/questions/40304985/server-side-events-sse-not-reaching-client-side
- it's recommended to use `webhooks` entry for describing SSE in OpenAPI 3.1 : https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#oas-webhooks
  - example: https://learn.openapis.org/examples/v3.1/webhook-example.html
- to make middlewares, I'd probably have to deal with mixins to avoid repeating EndpointsFactory implementation
  - https://www.typescriptlang.org/docs/handbook/mixins.html


RobinTail:
there should be probably another entity like `EventsFactory` holding a dictionary of established events

RobinTail:
I consider the experiment a success

src/sse.ts
```diff
@@ -0,0 +1,112 @@
...
+    .parse({ event, data });
+
+const headersTimeout = 1e4; // 10s to respond with a status code other than 200
+const ensureStream = (response: Response) =>
+  response.headersSent ||
+  response
+    .type(contentTypes.sse)
+    .setHeader("cache-control", "no-cache")
+    .setHeader("connection", "keep-alive")
+    .flushHeaders();
```
RobinTail, line 54: this method is not implemented by response mocking library.
Made an issue: https://github.com/eugef/node-mocks-http/issues/312
and a PR for simpler workaround: https://github.com/eugef/node-mocks-http/pull/313

CHANGELOG.md
```diff
@@ -2,6 +2,32 @@
 
 ## Version 21
 
+### v21.4.0
+
+- Feat: Introducing [Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events):
+  - Early implementation of event stream is now available using `unstable_createEventStream()` method;
+  - Client application can subscribe to the event stream using `EventSource` class instance;
+  - The functionality is limited: no middlewres yet possible.
```
RobinTail, line 10:
```suggestion
  - The functionality is limited: no middlewares possible yet.
```

src/server.ts
```diff
@@ -82,6 +82,7 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
   }
 
   const parsers: Parsers = {
+    sse: [],
```
RobinTail, line 85: maybe it should be partial

tests/unit/sse.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+  });
+
+  describe("ensureStream()", () => {
+    test("should set valid headers if they are not yet sent", () => {
+      const response = makeResponseMock();
+      ensureStream(response);
+      expect(response.statusCode).toBe(200);
+      // @todo enable when this one merged: https://github.com/eugef/node-mocks-http/issues/312
+      // expect(response.headersSent).toBeTruthy();
```
RobinTail, line 63: the PR is merged, but the fix is not yet released

RobinTail (commented):
reviewed

src/server.ts
RobinTail: done

tests/unit/sse.spec.ts
RobinTail: not critical, can do it later

RobinTail (commented):
ready

RobinTail:
Bug related to using `.flush()` #2347

PR #2239, opened 2024-12-08
Removing redundant type arguments on `Endpoint`

RobinTail:
- scopes and tags are only important for `BuildProps`, but not within `Endpoint` class
- I believe it's a rudimentary redundancy in the current arrangement
- This should ease making `output` schema optional


RobinTail (commented):
should be fine

PR #2240, opened 2024-12-08
Changing `BuildProps` type into interface

RobinTail:
Being type is also redundancy.
Should be interface after #2175


RobinTail (commented):
ok

PR #2245, opened 2024-12-09
Feat: `buildVoid` method

RobinTail:
Alternative to #2241 

This one should be better because it helps to avoid the situation when you just forget to place `output` schema while using `.build()`. The dedicated method. Not sure yet how to call it.
But it should be helpful for #2238

### DIFF
--- a/src/endpoints-factory.ts
+++ b/src/endpoints-factory.ts
@@ -18,7 +18,7 @@ import {
 
 interface BuildProps<
   IN extends IOSchema,
-  OUT extends IOSchema,
+  OUT extends IOSchema | z.ZodVoid,
   MIN extends IOSchema<"strip">,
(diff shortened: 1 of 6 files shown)


src/endpoints-factory.ts
```diff
@@ -147,6 +147,20 @@ export class EndpointsFactory<
       inputSchema: getFinalEndpointInputSchema<IN, BIN>(middlewares, input),
     });
   }
+
+  public buildEmpty<BIN extends IOSchema = EmptySchema>({
```
RobinTail, line 151: naming is questionable

RobinTail: ChatGPT advised `buildVoid()` and I liked it

RobinTail (commented):
nice

PR #2248, opened 2024-12-10
Fix `headersSent` issue when testing `writeHead()`

RobinTail:
Issue found while developing #2238 fixed in dependency


RobinTail (commented):
ok

PR #2251, opened 2024-12-11
Fix: no duplicates in `Path` type generated by `Integration`

RobinTail:
Found this one during my work on #2226


RobinTail (commented):
ok

PR #2258, opened 2024-12-20
Deprecating `jsonEndpoints` public const

RobinTail:
This is redundant and should be removed in v22.
The actual content-type header of the response should be used instead.


RobinTail (commented):
ok

PR #2266, opened 2024-12-24
Exp: uninstalling `tsx`

RobinTail:
This should deduplicate `esbuild`


RobinTail:
there is an issue of completing spawn process of `npx tsx` for some reason (ESM test)

PR #2267, opened 2024-12-25
Exp: Using native CORS mw lib

RobinTail:
- Benefits:
  - Should possibly make it easier to configure when using the well-known solution
- Disadvantages:
  - Last publish was 6 years ago: https://github.com/expressjs/cors/blob/master/lib/index.js
  - No custom headers option
  - more dependencies


RobinTail:
I don't like how that package `cors` is maintained.
It has substantial amount of issues that are not being solved for 6 years.
Therefore, I think, at least at the moment, I rather will keep the CORS implementation open for user customization.

PR #2274, opened 2024-12-31
ref(Integration): delegating props quotation to specific cases

RobinTail:
instead of `quoteProp()` helper there should be proper handling of particular cases

Should be helpful for #2269


src/integration.ts
```diff
@@ -285,13 +289,16 @@ export class Integration {
...
+              makePropertyIdentifier(propName),
+              f.createTrue(),
+            ),
           );
         }
         // "get /v1/user/retrieve": ["users"]
         endpointTags.push(
           f.createPropertyAssignment(
-            propName,
+            makePropertyIdentifier(propName),
```
RobinTail, line 301: extract

RobinTail (commented):
looking good

PR #2280, opened 2025-01-02
feat: `Subscription` class for SSE

RobinTail:
Addition to #2238


RobinTail:
I think it would be better to make it in v22

example/example.client.ts
```diff
@@ -446,6 +446,30 @@ export class ExpressZodAPIClient {
       ),
     );
   }
+  public subscribe<
+    K extends Extract<Request, `get ${string}`>,
+    R extends Extract<PositiveResponse[K], { event: string }>,
+  >(request: K, params: Input[K]) {
+    const path = request.split(/ (.+)/, 2)[1] as Path;
```
RobinTail, line 453: 1. this should probably be extracted into a private method, since it repeats implementation in `provide()`
2. there should also be possible path params substitution

src/integration-helpers.ts
```diff
@@ -65,12 +65,25 @@ export const makeEmptyInitializingConstructor = (
...
+  value: ts.TypeNode | ts.Identifier | string,
+) =>
   f.createPropertySignature(
     undefined,
     makePropertyIdentifier(name),
     undefined,
-    value,
+    typeof value === "string" || ts.isIdentifier(value)
+      ? f.createTypeReferenceNode(value)
+      : value,
```
RobinTail, line 78: this should be extracted into a low level helper

src/integration.ts
```diff
@@ -418,6 +427,158 @@ export class Integration {
...
+              ]),
+              f.createNumericLiteral(1),
+            ),
+            f.createTypeReferenceNode(this.ids.pathType),
+          ),
+        ),
+        makeConst(
+          this.ids.sourceConst,
+          makeNew(
+            f.createIdentifier("EventSource"),
```
RobinTail, line 456: this one is flagged in Node at the moment or only available via `undici` runtime.
constraints can not be established at the moment, but it's worth it to make a "todo" comment for the fu

src/integration-helpers.ts
RobinTail: e68c60fb

example/example.client.ts
RobinTail: extracted in 65cd7c25

RobinTail: proposed method
```ts
private substitute(path: string, params: Record<string, any>) {
    return Object.keys(params).reduce<[typeof path, typeof params]>(([final, remains], key) => {
      if (final.includes(`:${key}`)) final = final.replace(`:${key}`, params[key]);
      else remains[key] = params[key];
      return [final, remains];
    }, [path, {}]);
  }
```

RobinTail: alternative implementation without reducer having less types
```ts
private substitute(path: string, params: Record<string, any>) {
    const rest = {...params};
    for (const key in params)
      path = path.replace(`:${key}`, () => { delete rest[key]; return params[key] });
    return [path, rest] as const;
  }
```

RobinTail:
I think it needs Readme update

example/example.client.ts
```diff
@@ -422,29 +422,49 @@ export type Implementation = (
...
-            !path.includes(`:${key}`) && {
-              [key]: (params as Record<string, any>)[key],
-            },
-          ),
-        {},
-      ),
+    const [method, path] = this.parseRequest(request);
+    return this.implementation(method, ...this.substitute(path, params));
+  }
+  public subscribe<
```
RobinTail, line 445: maybe this should not be within client, because the client instantiation requires `Implementation` only needed for `provide()`, while `subscribe()` does not use it.
Perhaps `parseRequest()`, `substitute()` and `subscribe()` should be just functions.

example/example.client.ts
```diff
@@ -426,28 +440,37 @@ export class ExpressZodAPIClient {
...
 
+export const subscribe = <
+  K extends Extract<Request, `get ${string}`>,
+  R extends Extract<PositiveResponse[K], { event: string }>,
+>(
+  request: K,
+  params: Input[K],
+) => {
+  const [path, rest] = substitute(parseRequest(request)[1], params);
+  const source = new EventSource(
```
RobinTail, line 456: it may require `withCredentials: true`

example/example.client.ts
RobinTail: fixed in f66e1175

example/example.client.ts
RobinTail: perhaps can do it automatically by comparing `window.location.origin` to `new URL().origin`

example/example.client.ts
```diff
@@ -426,28 +440,37 @@ export class ExpressZodAPIClient {
...
+      );
+      return connection;
+    },
+  };
+  return connection;
+};
+
 // Usage example:
 /*
 export const exampleImplementation: Implementation = async (
```
RobinTail, line 476: this could become `defaultImplementation`

example/example.client.ts
```diff
@@ -426,28 +440,37 @@ export class ExpressZodAPIClient {
...
-          ),
-        {},
-      ),
-    );
+    const [method, path] = parseRequest(request);
+    return this.implementation(method, ...substitute(path, params));
   }
 }
 
+export const subscribe = <
```
RobinTail, line 448: maybe `class Subscription` with a potential for future customization?

example/example.client.ts
RobinTail: no, this is for cookies

example/example.client.ts
RobinTail: addressed in #2304

example/example.client.ts
RobinTail: should go after #2304

example/example.client.ts
```diff
@@ -469,4 +495,5 @@ export const exampleImplementation: Implementation = async (
 };
 const client = new ExpressZodAPIClient(exampleImplementation);
 client.provide("get /v1/user/retrieve", { id: "10" });
+client.subscribe("get /v1/events/time", {}).on("time", (time) => {});
```
RobinTail, line 498: usage is outdated

example/example.client.ts
RobinTail: made class in add0aa71

RobinTail:
should update readme and changelog

RobinTail (commented):
looks ready

RobinTail:
I'm happy with it now

PR #2299, opened 2025-01-11
ref(v22): Integration base

RobinTail:
Bricks for the building: this should make the `Integration` class a high level orchestrator of the codegen.

This should organize the code for readability.
Can cause minor conflicts for #2280


RobinTail:
I like the structure now

RobinTail (commented):
ok

PR #2303, opened 2025-01-13
ref(v22): `parseRequest()` and `substitute()` functions

RobinTail:
This should extract the refactoring and general improvement from #2280


RobinTail (commented):
nice

PR #2304, opened 2025-01-13
feat: Client default implementation

RobinTail:
This should come handy and make the client usage easier


example/example.client.ts
```diff
@@ -434,24 +434,7 @@ export type Implementation = (
...
-/*
-export const exampleImplementation: Implementation = async (
-  method,
-  path,
-  params,
-) => {
+const defaultImplementation: Implementation = async (method, path, params) => {
   const hasBody = !["get", "delete"].includes(method);
   const searchParams = hasBody ? "" : `?${new URLSearchParams(params)}`;
   const response = await fetch(
```
RobinTail, line 440: it might need CORS

RobinTail: helper introduced in 47724320

RobinTail: utilized in f2c97109

example/example.client.ts
```diff
@@ -428,14 +428,34 @@ const substitute = (path: string, params: Record<string, any>) => {
...
   path: string,
   params: Record<string, any>,
 ) => Promise<any>;
 
+const defaultImplementation: Implementation = async (method, path, params) => {
+  const hasBody = !["get", "delete"].includes(method);
+  const searchParams = hasBody ? "" : `?${new URLSearchParams(params)}`;
+  const url = new URL(`${path}${searchParams}`, "https://example.com");
+  const response = await fetch(url, {
+    mode: hasSameOrigin(url) ? "same-origin" : "cors",
```
RobinTail, line 444: it turned out it's default to `cors`

https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode
> Defaults to cors.

so, it's not needed

RobinTail: reverted in 5e871cb6

RobinTail (commented):
ok

PR #2307, opened 2025-01-13
ref(v22): Minimizing ids

RobinTail:
Some may be retrieved from the declarations of the corresponding variables.


RobinTail (commented):
nice

PR #2313, opened 2025-01-14
Renaming `ExpressZodAPIClient` to just `Client`

RobinTail:
1. Since I also wanna make subscription class in #2280 
2. In the era of modules there is no longer need to have a unique name


example/example.client.ts
```diff
@@ -434,7 +434,7 @@ export type Implementation = (
   params: Record<string, any>,
 ) => Promise<any>;
 
-export class ExpressZodAPIClient {
+export class Client {
```
RobinTail, line 437: needs migration

RobinTail (commented):
nice

PR #2324, opened 2025-01-18
feat: Pull examples from object schema properties

RobinTail:
When describing examples of the IO schemas only top level calls of `.example()` method used to be taken into account in the generated `Documentation`

```ts
factory.build({
  input: z
    .object({
      key: z.string().nonempty(),
    })
    .example({
      key: "1234-5678-90",
    }),
})
```

This feature enabled pulling examples up from the shape of object schema properties, so the syntax can be simplified:

```ts
factory.build({
  input: z.object({
    key: z.string().nonempty().example("1234-5678-90"),
  }),
})
```


example/example.documentation.yaml
```diff
@@ -118,9 +118,8 @@ paths:
           description: PATCH /v1/user/:id Parameter
           schema:
             type: string
-          examples:
-            example1:
-              value: "12"
+            examples:
+              - "12"
```
RobinTail, line 122: this change is not right, it's not visible in swagger UI

RobinTail: fixed in 38e03c8c

example/example.documentation.yaml
```diff
@@ -578,6 +590,8 @@ paths:
           description: GET /v1/events/time Parameter
           schema:
             type: string
+          examples:
+            example1: {}
```
RobinTail, line 594: where this comes from?

RobinTail: fixed in a11c68eb

src/common-helpers.ts
```diff
@@ -94,6 +94,18 @@ export const getExamples = <
   validate?: boolean;
 }): ReadonlyArray<V extends "parsed" ? z.output<T> : z.input<T>> => {
   const examples = schema._def[metaSymbol]?.examples || [];
+  if (schema instanceof z.ZodObject && !examples.length) {
+    const fromProps = Object.entries(schema.shape as z.ZodRawShape).reduce(
```
RobinTail, line 98: perhaps it should somehow go to `extractObjectSchema` to support more cases

RobinTail: not really, because it should also be applicable to response, which can be anything besides object

src/common-helpers.ts
RobinTail, line 1: there is call for the method in `arrayResultHandler` (used by `listUsersEndpoint`).
when pulling examples within endpoint down to the `items` level, they get duplicated because of that.
that should be avoided somehow

RobinTail: addressed in f4231d8d

RobinTail:
I want to revisit the composition of the established workflow:
- `getFinalEndpointInputSchema`
- `extractObjectSchema`
- `getExamples`

Perhaps something here should be simplified

RobinTail:
> * `getFinalEndpointInputSchema`
> * `extractObjectSchema`
> * `getExamples`
> Perhaps something here should be simplified


Reviewed. Once again, it *can* be simplified but for input flow only (moving copyMeta from getFinalEndpointInputSchema and pulling from getExamples into extractObjectSchema). 
The response part does not go through extractObjectSchema, because it can be not object-based.
The approach also complicates extractObjectSchema significantly.

src/documentation-helpers.ts
```diff
@@ -666,34 +666,38 @@ export const depictRequestParams = ({
...
+        composition === "components"
+          ? makeRef(paramSchema, depicted, makeCleanId(description, name))
+          : depicted;
+      return acc.concat({
+        name,
+        in: location,
+        required: !paramSchema.isOptional(),
+        description: depicted.description || description,
+        schema: result,
+        examples: depictParamExamples(extractObjectSchema(schema), name),
```
RobinTail, line 696: ⚠️ this one is duplicate of `objectSchema`

RobinTail: caused by d9cec4e57e1791fb50c0b96b4995ea9e8ef6c708

RobinTail: fixed in e7dcb779

src/common-helpers.ts
```diff
@@ -96,8 +112,21 @@ export const getExamples = <
...
+  /**
+   * @desc should pull examples from properties — applicable to ZodObject only
+   * @default true
+   * */
+  pullProps?: boolean;
 }): ReadonlyArray<V extends "parsed" ? z.output<T> : z.input<T>> => {
   const examples = schema._def[metaSymbol]?.examples || [];
+  if (!examples.length && pullProps && schema instanceof z.ZodObject) {
+    return getExamples({
+      schema: pullExampleProps(schema),
```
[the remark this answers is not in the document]
RobinTail: no, because the response workflow does not use it

src/common-helpers.ts
```diff
@@ -75,13 +75,29 @@ export const getMessageFromError = (error: Error): string => {
   return error.message;
 };
 
+/** Takes the original unvalidated examples from the properties of ZodObject schema shape */
+const pullExampleProps = <T extends z.SomeZodObject>(subject: T): T =>
```
RobinTail, line 79: can I reduce the complexity of this function?
it has 2 reducers and 2 external calls

src/common-helpers.ts
RobinTail, line 124: perhaps it could return just examples array to assign to `examples` here

src/common-helpers.ts
RobinTail: I reused `copyMeta` because I didn't want to repeat `combinations`.
However, this one is simpler, and `combinations` does not need `deepMergeRight`, because we combine examples from individual props that do not intersect.

src/common-helpers.ts
```diff
@@ -75,13 +75,26 @@ export const getMessageFromError = (error: Error): string => {
...
+  );
+
 export const getExamples = <
   T extends z.ZodTypeAny,
   V extends "original" | "parsed" | undefined,
 >({
   schema,
   variant = "original",
   validate = variant === "parsed",
+  pullProps = true,
```
RobinTail, line 97: perhaps this should not be default value.
can I find exact places where it should be activated?

src/documentation-helpers.ts
```diff
@@ -666,34 +671,38 @@ export const depictRequestParams = ({
...
-      in: location,
-      required: !shape[name].isOptional(),
-      description: depicted.description || description,
-      schema: result,
-      examples: depictParamExamples(schema, name),
-    });
-  }, []);
+  return Object.keys(objectSchema.shape).reduce<ParameterObject[]>(
+    (acc, name) => {
+      const paramSchema = objectSchema.shape[name];
```
RobinTail, line 676: this can be `.entries`

RobinTail (commented):
okay

PR #2327, opened 2025-01-20
Changing token header handling in example

RobinTail:
Perhaps should go after #2304 
Related to discussion #2326


RobinTail:
It turned out that `X-` prefix was deprecated in 2012.
https://www.rfc-editor.org/rfc/rfc6648.html

Which is even stated on MSDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

Not sure how I missed it doing #1183 , but it seems that we should make improvement towards opposite direction: keeping unprefixed `token` header and adjusting implementation to that.

RobinTail:
we shall go opposite way in v22 after doing #2332

PR #2328, opened 2025-01-21
feat: Optional context for `Client::provide()`

RobinTail:
See discussion #2326 

Featuring the ability to pass extra data into the `Implementation`.
It can be used for making a customized solution, in particular for dynamically assignable headers.


LucWag (approved):
Hey @RobinTail, great work. Looks good to me and many thanks for the implementation

RobinTail (commented):
now it looks good

PR #2332, opened 2025-01-22
ref: `LogicalContainer` handling

RobinTail:
1. I noticed a double mapping of logical containers with one more extra processing just in order to depict security schemas. I think the workflow could be simplified.
2. currently there is only a mapper and a single purpose processor, but if I make a generic walker out of it, it could be utilized for detecting non-prefixed headers for depictRequestParams


src/documentation-helpers.ts
```diff
@@ -9,8 +9,8 @@ import {
   ResponseObject,
   SchemaObject,
   SchemaObjectType,
-  SecurityRequirementObject,
   SecuritySchemeObject,
+  SecurityRequirementObject,
```
RobinTail, line 13:
```suggestion
    SecurityRequirementObject,
    SecuritySchemeObject,
```

src/documentation-helpers.ts
```diff
@@ -9,8 +9,8 @@ import {
   ResponseObject,
   SchemaObject,
   SchemaObjectType,
-  SecurityRequirementObject,
-  SecuritySchemeObject,
+    SecurityRequirementObject,
+    SecuritySchemeObject,
```
RobinTail, line 13:
```suggestion
  SecurityRequirementObject,
  SecuritySchemeObject,
```

RobinTail (commented):
brilliant!

PR #2333, opened 2025-01-22
`Endpoint::getSecurity()` to return an array of containers

RobinTail:
Cherry-picked from #2332 

This should ease that PR as later fix without any breaking changes


RobinTail (commented):
should be fine

PR #2336, opened 2025-01-23
feat: Configurable client class name

RobinTail:
I still wanna make the breaking change in v22 that changes the class name to just `Client`, but there will be a feature to configure it for consumer's needs


RobinTail (commented):
ok

PR #2337, opened 2025-01-23
feat(v22): Handle all headers

RobinTail:
Related to discussion: #2326 
- The idea is to handle all headers when enabled within `inputSources`
- The `Documentation` would need a way to distinguish them
  - Therefore a list of well-known headers will be included into the distribution
  - And there should probably be an option to include more

⚠️ The feature is risky: it may affect the behaviour for the users having `headers` enabled within `inputSources` with a high priority (in last place), because some header can be named exactly as another input.
Therefore I decided to move it to the next major (v22)


src/common-helpers.ts
```diff
@@ -49,7 +50,7 @@ export const getInput = (
     fallbackInputSource
   )
     .filter((src) => (src === "files" ? areFilesAvailable(req) : true))
-    .map((src) => (src === "headers" ? getCustomHeaders(req[src]) : req[src]))
+    .map((src) => (src === "headers" ? getHeaders(req[src]) : req[src]))
```
RobinTail, line 53: here we actually should take them all, no filtering needed

src/common-helpers.ts
```diff
@@ -61,7 +54,7 @@ export const getInput = (
     fallbackInputSource
   )
     .filter((src) => (src === "files" ? areFilesAvailable(req) : true))
-    .map((src) => (src === "headers" ? getCustomHeaders(req[src]) : req[src]))
+    .map((src) => req[src])
     .reduce<FlatObject>((agg, obj) => Object.assign(agg, obj), {});
```
RobinTail, line 58: perhaps could combine them

RobinTail (commented):
ready

PR #2341, opened 2025-01-25
Fixing diagram edits and renaming Middleware public prop

RobinTail:
- i noticed it became not ediable
- `middleware` was renamed to `handler` in `Middleware`


RobinTail (commented):
ok

PR #2351, opened 2025-01-29
Improving Typescript API

RobinTail:


eslint.config.js
```diff
@@ -102,8 +100,11 @@ const tsFactoryConcerns = [
     message: "use makePromise() helper",
   },
   {
-    selector:
-      "CallExpression[callee.property.name='createTypeReferenceNode'][arguments.length=1]",
+    selector: "Identifier[name='createTypeReferenceNode']",
+    message: "use ensureTypeNode() helper",
+  },
+  {
+    selector: "Identifier[name='createKeywordTypeNode']",
```
RobinTail, line 107: can combine them in a regex

src/typescript-api.ts
```diff
@@ -269,42 +272,37 @@ export const makeInterface = (
...
+  ).map(([name, val]) => {
+    const { type, init } =
+      typeof val === "object" && "init" in val ? val : { type: val };
     return f.createTypeParameterDeclaration(
       [],
       name,
-      type && ensureTypeNode(type),
-      typeof init === "number" ? f.createKeywordTypeNode(init) : init,
+      type ? ensureTypeNode(type) : undefined,
+      init ? ensureTypeNode(init) : undefined,
```
RobinTail, line 291: that could be genetic to return undefined too, then expression can be shorter

RobinTail (commented):
there is no limit to perfection, therefore i think it's good enough for now

PR #2366, opened 2025-01-31
feat: 405: Method not allowed

RobinTail:
There is an HTTP code 405 for routes that exist, but require different method.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

Unfortunately, `req.route` is `undefined` when within middlewares, so have to use `app.all()` as recommended.


src/server.ts
```diff
@@ -93,6 +93,8 @@ export const createServer = async (config: ServerConfig, routing: Routing) => {
   initRouting({ app, routing, getLogger, config, parsers });
   app.use(parserFailureHandler, notFoundHandler);
 
+  logger.info("stack", app.router.stack);
+
```
RobinTail, line 97:
```suggestion
```

src/server-helpers.ts
```diff
@@ -44,13 +57,40 @@ export const createParserFailureHandler =
...
-      404,
-      `Can not ${request.method} ${request.path}`,
+    const supportedMethods = findSupportedMethods(
+      request.path,
+      request.app?.router.stack,
     );
+    const error = supportedMethods?.length
+      ? createHttpError(405, `${request.method} is not allowed`, {
+          headers: { Allowed: supportedMethods.join(", ") },
+        })
```
RobinTail, line 92: maybe! it should be under a flag

src/server-helpers.ts
```diff
@@ -44,13 +57,43 @@ export const createParserFailureHandler =
...
+    getLogger,
+    config: { wrongMethodBehavior = 404 },
+  }: HandlerCreatorParams & { config: CommonConfig }): RequestHandler =>
   async (request, response) => {
-    const error = createHttpError(
-      404,
-      `Can not ${request.method} ${request.path}`,
-    );
+    const supportedMethods =
+      wrongMethodBehavior === 405 && request.app?.router.stack
```
RobinTail, line 89: `app.router` is not present in express v4

RobinTail: in this regard, should rather use own methods than hacking express.router

RobinTail: fixed

src/routing.ts
```diff
@@ -56,11 +57,26 @@ export const initRouting = ({
...
+        app.options(path, ...matchingParsers, handler);
+        familiar.get(path)?.push("options");
+      }
     }
+    familiar.get(path)?.push(method);
     app[method](path, ...matchingParsers, handler);
   };
   walkRouting({ routing, onEndpoint, onStatic: app.use.bind(app) });
+  if (config.wrongMethodBehavior !== 405) return;
+  for (const [path, allowedMethods] of familiar.entries()) {
```
RobinTail, line 72: need unit test for it

src/routing.ts
```diff
@@ -56,11 +57,26 @@ export const initRouting = ({
...
+    app.all(path, ({ method }, res, next) => {
+      return next(
```
RobinTail, line 74: this is handled by `createParserFailureHandler`.
perhaps it should be generalized

src/routing.ts
```diff
@@ -16,6 +17,15 @@ export interface Routing {
 
 export type Parsers = Partial<Record<ContentType, RequestHandler[]>>;
 
+export const createWrongMethodHandler =
+  (allowedMethods: Array<Method | AuxMethod>): RequestHandler =>
+  ({ method }, _res, next) =>
+    next(
+      createHttpError(405, `${method} is not allowed`, {
+        headers: { Allowed: allowedMethods.join(", ").toUpperCase() },
```
RobinTail, line 25: what if there is a custom ResultHandler that does not care about setting headers from error?

RobinTail (commented):
ok

RobinTail (commented):
now ok

PR #2377, opened 2025-02-09
Feat: checking the acceptance of path params by assigned endpoint

RobinTail:
historically, there is a technical gap between `Routing` and `Endpoint`.
There are no constraints between route path params and input schema that is (not?) using them.

Bench results on using `in`

```
 ✓ bench/experiment.bench.ts > Experiment for key lookup 10192ms
     name                               hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · in                      15,462,722.86  0.0001  0.0639  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.18%  7731362   fastest
   · R.has                    2,298,774.34  0.0004  0.0562  0.0004  0.0004  0.0005  0.0006  0.0008  ±0.15%  1149388   slowest
   · Object.keys + includes  10,066,568.65  0.0001  0.4847  0.0001  0.0001  0.0001  0.0001  0.0003  ±1.17%  5033285

 BENCH  Summary

  in - bench/experiment.bench.ts > Experiment for key lookup
    1.54x faster than Object.keys + includes
    6.73x faster than R.has
```


express-zod-api/src/diagnostics.ts
```diff
@@ -2,40 +2,72 @@ import { responseVariants } from "./api-response";
 import { FlatObject } from "./common-helpers";
 import { contentTypes } from "./content-type";
 import { assertJsonCompatible } from "./deep-checks";
+import { getRoutePathParams } from "./documentation-helpers";
```
RobinTail, line 5: should be moved to common helpers

express-zod-api/src/diagnostics.ts
```diff
@@ -2,40 +2,72 @@ import { responseVariants } from "./api-response";
...
+        if (issue) {
+          this.logger.warn(
+            `The endpoint assigned to ${path} probably does not accept its path parameter ${param}`,
+            Object.assign(ctx, { sample, issue }),
+          );
+        }
+      } else if (
+        typeof result.data === "object" &&
+        result.data !== null &&
+        !(param in result.data)
```
RobinTail, line 64: it can be false-positive in case of top level transformations and `.remap()`

express-zod-api/src/diagnostics.ts
```diff
@@ -1,41 +1,72 @@
...
+    path: string,
+    endpoint: AbstractEndpoint,
+    ctx: FlatObject,
+  ): void {
+    const params = getRoutePathParams(path);
+    for (const param of params) {
+      const sample = { [param]: "123" };
+      const result = endpoint.getSchema("input").safeParse(sample);
+      if (!result.success) {
+        const issue = result.error.issues.find(
```
RobinTail, line 51: this does not work when param name differs from the input.
we should do opposite, probably using `extractObjectSchema()`

express-zod-api/src/documentation-helpers.ts
RobinTail, line 1: need to reduce diff here

express-zod-api/src/diagnostics.ts
```diff
@@ -1,41 +1,58 @@
...
         }
       }
-      this.#verified.add(endpoint);
+    }
+    this.#verifiedEndpoints.add(endpoint);
+  }
+
+  public checkPathParams(
+    path: string,
+    endpoint: AbstractEndpoint,
```
RobinTail, line 44: should probably avoid checking same combination for another method

express-zod-api/src/diagnostics.ts
RobinTail, line 1: should make some tests

RobinTail (commented):
I'm happy with it

PR #2384, opened 2025-02-10
Better compatibility test

RobinTail:
related to #2382


RobinTail (commented):
Good for now. Then I will make it in Docker.

PR #2390, opened 2025-02-11
feat: Deprecations

RobinTail:
Based on #2389

PR #2392, opened 2025-02-12
Fix `Metadata` constraints

RobinTail:
- using `ZodTypeAny` makes `_def: any`.
- found it during #2390 
- should rather use `ZodType` where possible
- `ZodRawShape` always refers to `ZodTypeAny` so have to use `as ZodType` statement
- several places using `_def.typeName` from `ZodTypeAny` which works because of `any`


RobinTail (commented):
ok

PR #2398, opened 2025-02-13
ref(v23): Using getters

RobinTail:
Todo from #2390


express-zod-api/src/endpoint.ts
```diff
@@ -127,19 +129,19 @@ export class Endpoint<
     );
   }
 
-  public override getSecurity() {
+  public override get security() {
     return Object.freeze(
       (this.#def.middlewares || [])
-        .map((middleware) => middleware.getSecurity())
+        .map((middleware) => middleware.security)
         .filter((entry) => entry !== undefined),
```
RobinTail, line 136: this can be simplified with R.reject and R.pluck/R.prop

RobinTail (commented):
awesome

PR #2399, opened 2025-02-13
Revert "Fix: `Endpoint::getSecurity()` returns `ReadonlyArray`"

RobinTail:
Reverts RobinTail/express-zod-api#2396

After checking a little, I realized that it's not really needed, because the returned array has no reference to internal property

### DIFF
--- a/express-zod-api/tests/endpoint.spec.ts
+++ b/express-zod-api/tests/endpoint.spec.ts
@@ -318,27 +318,6 @@ describe("Endpoint", () => {
     );
   });
 
-  describe(".getSecurity()", () => {
-    test("should return a readonly array of security based logical containers", () => {
-      const endpoint = defaultEndpointsFactory
-        .addMiddleware({
-          security: { type: "header", name: "X-Token" },
-          handler: vi.fn(),
-        })
-        .addMiddleware({
-          security: { type: "header", name: "X-API-Key" },
-          handler: vi.fn(),
-        })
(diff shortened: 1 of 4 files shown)


express-zod-api/tests/endpoint.spec.ts
```diff
@@ -318,27 +318,6 @@ describe("Endpoint", () => {
...
-      const result = endpoint.getSecurity();
-      expect(result).toEqual([
-        { name: "X-Token", type: "header" },
-        { name: "X-API-Key", type: "header" },
-      ]);
-      expect(() => (result as any[]).push()).toThrowError(/read only/);
-    });
-  });
-
   describe("getRequestType()", () => {
```
RobinTail, line 321:
```suggestion
  describe(".getSecurity()", () => {
    test("should return a array of security based logical containers", () => {
      const endpoint = defaultEndpointsFactory
        .addMiddleware({
          security: { type: "header", name: "X-Token" },
          handler: vi.fn(),
        })
        .addMiddleware({
          security: { type: "header", name: "X-API-Key" },
          handler: vi.fn(),
        })
        .build({ output: z.object({}), handler: vi.fn() });
      const result = endpoint.getSecurity();
      expect(result).toEqual([
        { name: "X-Token", type: "header" },
        { name: "X-API-Key", type: "header" },
      ]);
    });
  });

  describe("getRequestType()", () => {
```

RobinTail (commented):
should be safe enough

PR #2401, opened 2025-02-14
ref(v23): Undisclosed internals

RobinTail:
I no longer want to consider changes to the internal props/method as breaking changes, therefore I'm gonna mark them as `@internal` and exclude from DTS.


RobinTail (commented):
good

PR #2402, opened 2025-02-14
ref(v23): More internals

RobinTail:
Hiding more entities, similar to #2401 

This should also reduce the DTS size, removing entities that are not supposed to be used by consumer


RobinTail:
for good

PR #2410, opened 2025-02-15
Using namespace import from `ramda`

RobinTail:
- no multiline import in case of multiple functions needed
- clarity
- no shadow const concerns
- can use all functions I need

### DIFF
--- a/express-zod-api/src/zts.ts
+++ b/express-zod-api/src/zts.ts
@@ -1,4 +1,4 @@
-import { chain, eqBy, path, prop, tryCatch, uniqWith } from "ramda";
+import * as R from "ramda";
 import ts from "typescript";
 import { z } from "zod";
 import { hasCoercion, getTransformedType } from "./common-helpers";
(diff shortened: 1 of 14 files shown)


express-zod-api/src/zts.ts
```diff
@@ -150,13 +150,16 @@ const onRecord: Producer = (
...
-    const uniqs = uniqWith((...props) => {
-      if (!eqBy(nodePath.name, ...props)) return false;
-      if (eqBy(nodePath.type, ...props) && eqBy(nodePath.optional, ...props))
+    const members = R.chain(R.prop("members"), nodes);
+    const uniqs = R.uniqWith((...props) => {
+      if (!R.eqBy(nodePath.name, ...props)) return false;
+      if (
+        R.eqBy(nodePath.type, ...props) &&
+        R.eqBy(nodePath.optional, ...props)
+      )
```
RobinTail, line 162: R.both?

RobinTail (commented):
ok

PR #2419, opened 2025-02-18
Using allowed dependencies plugin instead of unicorn

RobinTail:
Alternative to #2417


RobinTail:
Here will make handling for devDependencies, so that could use it not only for sources
https://github.com/RobinTail/allowed-dependencies/pull/40

yarn.lock
RobinTail, line 1: amazing!

RobinTail (commented):
good

PR #2424, opened 2025-02-20
Organizing routing entities

RobinTail:


express-zod-api/src/depends-on-method.ts
```diff
@@ -1,7 +1,8 @@
 import * as R from "ramda";
 import { AbstractEndpoint } from "./endpoint";
 import { Method } from "./method";
-import { Routable } from "./routable";
+
```
RobinTail, line 4:
```suggestion
```

RobinTail:
could not move walkRouting to Routing — causes circular referenced between files for esbuild

PR #2374, opened 2025-02-07
Ref: Workspaces

RobinTail:
Reiteration of #1351 having improved dependencies and environment

PR #2067, opened 2024-10-01
Exp: Trying `pnpm`

RobinTail:
https://dev.to/andreychernykh/yarn-npm-to-pnpm-migration-guide-2n04

PR #2427, opened 2025-02-21
Exp: trying `yarn@v4`

RobinTail:
- improved performance
- modern implementation
- better support for workspaces
  - complements #2374 
  - [X] fixed peer warnings among workspaces
  - [X] fixes issue with `attw --pack` ✅ 
  - [X] dependencies can be listed correctly for all workspaces including references using `workspace:^`
    - [ ] `eslint-plugin-allowed-dependencies` should be reconfigured more precisely for each workspace — address
- has `yarn dedupe`, instead of `npx yarn-deduplicate`
- [X] `prebuild` and `postbuild` commands are no longer supported — should be addressed fully
- ❌  `yarn upgrade` removed
  - `yarn up [-R]` — is not good alternative
  - [x] consider https://github.com/antfu-collective/taze
    - `yarn workspace express-zod-api dlx taze -w -i`
- echo of commands was removed https://github.com/yarnpkg/berry/issues/1215
  - [X] restoring it using the plugin https://github.com/leaumar/yarn-plugin-echo-execute
  - plugin builder documentation: https://github.com/yarnpkg/berry/blob/HEAD/packages/yarnpkg-builder/README.md
- [ ] `--preid` was removed on `yarn version`, can not do `beta`


RobinTail:
- The problem with `yarn version` having no way to configure `--preid` is significant.
- `yarn up` is not good, though `taze` is ok
- `yarn publish` still does not have `--dry-run`
Should rather retry #2067 which was cancelled due to the need of workspaces, that I do have now

RobinTail:
wait a second... I've got an idea...

RobinTail:
no

PR #2450, opened 2025-02-26
Exp: on `attw`

RobinTail:
The weird `postbuild` command was introduced in https://github.com/RobinTail/express-zod-api/pull/2374 in response `attw` failure to work together with `npm publish -w` (within workspace), see https://github.com/RobinTail/express-zod-api/pull/2374/commits/1b556540c32781852be0d5a9018835a1463710c0

### DIFF
--- a/express-zod-api/package.json
+++ b/express-zod-api/package.json
@@ -17,7 +17,7 @@
   "funding": "https://github.com/sponsors/RobinTail",
   "scripts": {
     "build": "tsup",
-    "postbuild": "yarn pack --filename release.tgz && attw release.tgz && rm release.tgz",
+    "postbuild": "attw --pack",
     "pretest": "tsc --noEmit",
     "test": "vitest run --coverage",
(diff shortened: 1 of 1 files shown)


RobinTail:
I believe running `npm pack` (which is what `attw --pack` doing) is not possible in the context of running `npm publish`

RobinTail:
Yep. This hook fails for example when running `npm publish --dry-run`

```
"prepublishOnly": "npm pack && ls -la *.tgz"
```

because `npm pack` does not really produce the packed file

RobinTail:
so `postbuild` with `attw --pack` is ok for build, but the sequence within `prepublishOnly` should be different

RobinTail:
> publish does not save the tarball locally

https://docs.npmjs.com/cli/v9/using-npm/scripts#npm-publish

RobinTail:
I will have to switch either to #2427 or #pnpm

PR #2428, opened 2025-02-21
Exp: trying `pnpm` again

RobinTail:
Reiterating #2067 
Based on the findings of #2427 
And using the advantage of having workspaces introduced in #2374

PR #2451, opened 2025-02-26
Exp: `bun`

RobinTail:
Alternative to #2428 and #2427 

Issues
- snyk seems to be not supporting bun
- there is no `bun why` yet, but turned out can use `npm why` and it works
- no `bun version` but can use `bumpp`
- can remove `tsx`


RobinTail:
`bun update` mutates `peerDependencies` even with `--omit=peer`
`¯\_(ツ)_/¯`

PR #2475, opened 2025-03-19
Drop Express 4

RobinTail:
Express 5 provides better support for async handlers.
https://expressjs.com/en/guide/writing-middleware.html

It automatically handles rejected `Promise` returned from a Middleware, so that there is no need to call `next(error)`.

v5.1 is also finally marked `latest`


RobinTail:
- [ ] should also remove `depd` — removed in 5.1
  - https://github.com/expressjs/express/pull/6197/files

RobinTail (commented):
ok

PR #2357, opened 2025-01-30
v23 is for Sonia

RobinTail:
image

Sonia Rescalvo Zafra was a Spanish trans woman who was murdered by a group of neo-Nazis in the _Parc de la Ciutadella_ in Barcelona on 6 October 1991. Her murder is notable for being the first known case in Spain of a person being murdered for being transgender.

Sonia was born in _Cuenca_, Spain in 1946. She moved to Barcelona at the age of 16 and began working as a showgirl in theatres on the _Avinguda del Paral·lel_. Later in life, she fell on hard times and became homeless and addicted to drugs, turning to sex work to survive.

Sonia and a her friend, Doris Romero, were sleeping rough at a bandstand in Barcelona's Parc _de la Ciutadella_ in the early hours of 6 October 1991 when they were abruptly attacked by a group of six teenage skinheads from the _Boixos Nois_, who beat them with a broomstick and repeatedly kicked them in the head while wearing steel-toe boots. Sonia was killed in the attack, while Doris was seriously injured and left unconscious. The perpetrators went on to attack three other people sleeping near the bandstand. One of the three other victims, Miguel Pérez Barreiros, was able to contact police after the attack, despite being left completely blind from his injuries. When police discovered Sonia's body, she was so badly bruised that media reports initially misidentified her as a black woman.

In June 1994, six of the suspects were convicted of participating in the attack and given prison sentences ranging from 23 to 50 years, while the seventh, Lozano, was fined 100,000 pesetas for knowing of the murder and not reporting it. In 1996, the attackers' sentences were reduced by the Supreme Court of Spain. All six have since been released from prison.

https://en.wikipedia.org/wiki/Murder_of_Sonia_Rescalvo_Zafra

Transgender women suffer too frequently from transphobic violence and cruelty, being the less protected social group. I'd like to raise an awareness of this problem. Humans should be creators — not killers. But most importantly, I want every transgender girl to have an opportunity to create applications quickly and, in general, learn to write code easily in order to receive job offers and leave dangerously transphobic territories for more favorable and civilized ones, and live happily there. Protect transgender women.

---------------------------------

[...]

PR #2476, opened 2025-03-19
Deprecating v18

RobinTail:
Planned for April, probably before #2357


RobinTail (commented):
deprecated v18 on NPM

PR #2494, opened 2025-03-27
Exp: Async logging

RobinTail:
This should be a major performance boost for the `BuiltinLogger`
https://expressjs.com/en/advanced/best-practice-performance.html#do-logging-correctly


RobinTail:
wrong base

PR #2495, opened 2025-03-27
Exp: Async logging

RobinTail:
This should be a major performance boost for the `BuiltinLogger`
https://expressjs.com/en/advanced/best-practice-performance.html#do-logging-correctly

```
     name             hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · sync     248,537.46  0.0034  1.1198  0.0040  0.0037  0.0062  0.0078  0.0301  ±1.66%   124269
   · async  6,109,895.93  0.0001  1.0178  0.0002  0.0001  0.0020  0.0021  0.0026  ±0.83%  3055099   fastest
```

Problems:
- maintaining queue ✅ 
  - `setImmediate()` seems to be doing it good enough, but:
- purging the queue on termination ✅ 
    - using a proprietary stack and use `setImmediate()` only for printing from it
- collaboration with a graceful shutdown ✅ 
  - modified `installTerminationListener()`
- child loggers ✅  
   - introducing a registry and the public `dispose()` method

### DIFF
--- a/express-zod-api/src/builtin-logger.ts
+++ b/express-zod-api/src/builtin-logger.ts
@@ -95,7 +95,7 @@ export class BuiltinLogger implements AbstractLogger {
 
   protected postpone(line: string) {
     this.stack.push(line);
-    setImmediate(this.purge.bind(this));
+    setImmediate(this.purge.bind(this)).unref(); // process.exit has higher priority
   }
 
   /** @internal */
--- a/express-zod-api/src/builtin-logger.ts
+++ b/express-zod-api/src/builtin-logger.ts
@@ -98,7 +98,8 @@ export class BuiltinLogger implements AbstractLogger {
     setImmediate(this.purge.bind(this));
   }
 
-  protected purge() {
+  /** @internal */
+  public purge() {
     while (this.stack.length) console.log(this.stack.shift());
   }
 
--- a/express-zod-api/src/builtin-logger.ts
+++ b/express-zod-api/src/builtin-logger.ts
@@ -100,15 +100,18 @@ export class BuiltinLogger implements AbstractLogger {
     this.postponed = setImmediate(this.purge.bind(this)).unref(); // do not block process.exit()
   }
 
-  /** @internal */
-  public purge() {
+  protected purge() {
     while (this.stack.length) console.log(this.stack.shift());
   }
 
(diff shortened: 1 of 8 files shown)


express-zod-api/src/builtin-logger.ts
```diff
@@ -82,7 +91,31 @@ export class BuiltinLogger implements AbstractLogger {
...
+    while (this.stack.length) console.log(this.stack.shift());
+  }
+
+  /**
+   * @internal
+   * @desc Prepares the instance for deleting. Callback is for postponed deletion.
+   * */
+  public dispose(cb?: () => void) {
+    this.config.level = "silent"; // no more logs
+    clearImmediate(this.postponed);
```
RobinTail, line 113: make early exit if it's not async

express-zod-api/src/server-helpers.ts
```diff
@@ -153,16 +160,27 @@ export const installDeprecationListener = (logger: ActualLogger) =>
...
-  options: { timeout, events = ["SIGINT", "SIGTERM"] },
-}: {
-  servers: Parameters<typeof monitor>[0];
-  options: Extract<ServerConfig["gracefulShutdown"], object>;
-  logger: ActualLogger;
-}) => {
-  const graceful = monitor(servers, { logger, timeout });
-  const onTerm = () => graceful.shutdown().then(() => process.exit());
+export const installTerminationListener = (
+  servers: Array<http.Server | https.Server>, // @todo extract
```
RobinTail, line 164: extract the type

express-zod-api/src/server.ts
```diff
@@ -31,19 +31,28 @@ const makeCommonEntities = (config: CommonConfig) => {
   const logger = isLoggerInstance(config.logger)
     ? config.logger
     : new BuiltinLogger(config.logger);
+  const childLoggers =
+    logger instanceof BuiltinLogger && config.childLoggerProvider
+      ? new Set<BuiltinLogger>()
```
RobinTail, line 36: mayabe `DisposableStack`?

RobinTail:
this implementation does not make a dramatic difference, because the same thread is used for printing the same about of bytes to the sync stream of `stdout`.
Instead, should try dedicated worker thread in another PR

PR #2522, opened 2025-04-06
Using `errorHandler` from config for `testMiddleware()`

RobinTail:
To reduce the confusion, using `errorHandler` which is already present in config.
This is breaking.
Also, `testMiddleware` will not throw now.

### DIFF
--- a/express-zod-api/src/testing.ts
+++ b/express-zod-api/src/testing.ts
@@ -2,8 +2,6 @@ import { Request, Response } from "express";
 import { ensureError, FlatObject, getInput } from "./common-helpers";
 import { CommonConfig } from "./config-type";
 import { AbstractEndpoint } from "./endpoint";
-import { ResultHandlerError } from "./errors";
-import { lastResortHandler } from "./last-resort";
 import {
   AbstractLogger,
   ActualLogger,
@@ -148,15 +146,11 @@ export const testMiddleware = async <
     const output = await middleware.execute(commons);
     return { requestMock, responseMock, loggerMock, output };
   } catch (e) {
-    const error = ensureError(e);
-    try {
-      await errorHandler.execute({ ...commons, error, output: null });
-    } catch (handlingError) {
-      lastResortHandler({
-        ...commons,
-        error: new ResultHandlerError(error, ensureError(handlingError)),
-      });
-    }
+    await errorHandler.execute({
+      ...commons,
+      error: ensureError(e),
+      output: null,
+    });
     return { requestMock, responseMock, loggerMock, output: {} };
   }
 };
--- a/express-zod-api/src/testing.ts
+++ b/express-zod-api/src/testing.ts
@@ -2,6 +2,8 @@ import { Request, Response } from "express";
 import { ensureError, FlatObject, getInput } from "./common-helpers";
 import { CommonConfig } from "./config-type";
 import { AbstractEndpoint } from "./endpoint";
+import { ResultHandlerError } from "./errors";
+import { lastResortHandler } from "./last-resort";
 import {
   AbstractLogger,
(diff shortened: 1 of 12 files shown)


RobinTail:
this needs a migration

express-zod-api/src/testing.ts
```diff
@@ -120,34 +123,40 @@ export const testMiddleware = async <
...
+  } catch (e) {
+    const error = ensureError(e);
+    try {
+      await errorHandler.execute({ ...commons, error, output: null });
+    } catch (handlingError) {
+      lastResortHandler({
+        ...commons,
+        error: new ResultHandlerError(error, ensureError(handlingError)),
+      });
+    }
```
RobinTail, line 159: all that actually replicates the workflow established in `Endpoint::execute()`

Perhaps, it would be possible to reuse it instead.

RobinTail: tried it, but it's getting not so neat actually, executes a lot of other stuff besides the subject and requires to extract clean output

RobinTail: no, it actually could be interesting and more similar to how `testEndpoint()` works...

RobinTail: sample solution

```ts
export const testMiddleware = async <
  LOG extends FlatObject,
  REQ extends RequestOptions,
  OPT extends FlatObject,
  OUT extends FlatObject,
  IN extends IOSchema<"strip">,
>({
  middleware,
  options = {} as OPT,
  ...rest
}: TestingProps<REQ, LOG> & {
  /** @desc The middleware to test */
  middleware: Middleware<OPT, OUT, string, IN>;
  /** @desc The aggregated output from previously executed middlewares */
  options?: OPT;
}) => {
  const { requestMock, responseMock, loggerMock, configMock } =
    makeTestingMocks(rest);

  let output = {};

  await new EndpointsFactory(configMock.errorHandler ?? defaultResultHandler)
    .addMiddleware({ handler: async () => options })
    .addMiddleware({
      input: middleware.schema,
      handler: async (params) => {
        output = await middleware.execute(params);
        return output;
      },
    })
    .buildVoid({ handler: async () => {} })
    .execute({
      request: requestMock,
      response: responseMock,
      logger: loggerMock,
      config: configMock,
    });

  return { requestMock, responseMock, loggerMock, output };
};
```

RobinTail: simpler one, preserving `AbstractMiddleware`:

```ts
export const testMiddleware = async <
  LOG extends FlatObject,
  REQ extends RequestOptions,
>({
  middleware,
  options = {},
  ...rest
}: TestingProps<REQ, LOG> & {
  /** @desc The middleware to test */
  middleware: AbstractMiddleware;
  /** @desc The aggregated output from previously executed middlewares */
  options?: FlatObject;
}) => {
  const { requestMock, responseMock, loggerMock, configMock } =
    makeTestingMocks(rest);

  let output = {};

  await new EndpointsFactory(configMock.errorHandler ?? defaultResultHandler)
    .addMiddleware({ handler: async () => options })
    .addMiddleware({
      input: middleware.schema,
      handler: async (params) => {
        output = await middleware.execute(params);
        return output;
      },
    })
    .buildVoid({ handler: async () => {} })
    .execute({
      request: requestMock,
      response: responseMock,
      logger: loggerMock,
      config: configMock,
    });

  return { requestMock, responseMock, loggerMock, output };
};
```

RobinTail: all that unfortunately does not work well for SSE, where Endpoint is not supposed to finish early.

RobinTail (commented):
ok

PR #2525, opened 2025-04-07
Aligning catcher with documentation

RobinTail:
Removing the status code coercion to `400` in the catcher, so that errors other than `HttpError` lead to `500` response by default.
Found the discrepancy from the documented behaviour on error handling.
Most likely introduced in v19.0.0 by #1741 

Pick from the documentation

> - Ones related to routing, parsing and upload issues — handled by `ResultHandler` assigned to `errorHandler` in config:
>   - Default is `defaultResultHandler` — it sets the response status code from the corresponding `HttpError`:
>     `400` for parsing ⚠️ 👀   **(this should be corrected to what specific parser throws)**, `404` for routing ✅ ,  `config.upload.limitError.statusCode` for upload issues ✅ , or `500` for others ⚠️  **(this one was 400 in catcher, fixed below)**;

PR #2526, opened 2025-04-07
Fix for `ez.raw()` without extension

RobinTail:
Noticed that its output type does not work properly when no extension argument given.
The issue seems to be introduced in v19.0.0
Discovered and needed for #2525


RobinTail (commented):
ok

PR #2537, opened 2025-04-10
Migration to Zod 4

RobinTail:
This PR is for migrating the framework onto Zod 4 as is and preserving its original behaviour and functionality as much as possible, except only the utilization of the new Metadata system for its purposes.
All other improvements, taking advantages of new features, simplifications and prettifying should be separated into dedicated PRs.

The current problems:
- IOSchema ✅ 
  - incompability of zod 4 schemas between each other, in particular Typescript does not like that certain schemas have different `brand()` and `_zod.values`
  - is inability of Typescript to handle circular nature of the `IOSchema` type when it's based on `ZodObject` from Zod 4.
  - the cause seems to be top level transformations chain
  - possible simplification could be any schema that leads to an object, but not array: `z.ZodType<object & { [Symbol.iterator]?: never }>`
  - made a runtime validation
- Plugin ✅ 
  - ZodType is not patchable through prototypes
  - found a workaround by patching each `Zod*` class
  - however, since meta system is now detached, it won't withstand refinements
  - found a workaround by taking advantage of setter
- Traversing schemas ✅ 
  - Schema of `zod` seem to contain and refer to `@zod/core` as children, not other `zod` ones
  - Found a way to rewrite rules based on `@zod/core`
- Iterating checks 🆗  
  - check type is `string`, not literal, have to introduce a lot of `satisfies`
  - they all have base-type `never` for some reason, that makes them unusable for `.find()`
  - checks can also contain other schemas `ZodISODate` for example which is not typed at all
  - `z.base64()` is not the same as `z.string().base64()`, but they both have `type: string`, so the check itself can act as a schema, and it should be handled accordingly
    - made handling for those cases
  - ⚠️  currently things like `z.int().max(100)` do not work due to external bug, so it's unclear what exactly such definition would look like
- `ZodError` does not inherit `Error` anymore 🆗  
  - this should require an extra care to revisit all the code
  - ~~`issues` property marked as internal and deprecated, probably should use `z.treeifyError`~~
    - that note was [removed in core 0.5.1](https://github.com/colinhacks/zod/pull/4074/commits/ee6b5f419594b3fecae68cabf23eaf3df822a3b6#diff-3c955f0f8d911dbf3b8523f50874b91753fecdf8d4c020087761e2807e895d56L178)
  - `ensureError()` will make an error out of `ZodError` with message
- `ZodLazy` now has a `getter()`  ✅ 
  - that always returns a new schema instance, which is not deduplicatble by `Map`
  - fixed by mapping over the getter
- Optionals are buggy ✅ 
  -  `_zod.qin` is not set on `ZodOptional`
  - `isOptional()` may throw despite using `safeParse` when it's dealing with `ZodPromise`
  - handled all that


RobinTail:
The approach on `IOSchema`, its nature and using `ZodIntersecion` should probably be changed in order to adopt all that
https://v4.zod.dev/v4/changelog#zintersection

RobinTail:
Interesting workaround on limited recursions, but I won't go that way

```ts
/**
 * @link https://dev.to/scooperdev/supporting-circularly-referenced-mapped-types-in-typescript-4825
 */

type Depth = [1, 2, 3, never];
export type Inc<T> = T extends number ? Depth[T] : never;
```

express-zod-api/tests/documentation-helpers.spec.ts
```diff
@@ -446,11 +462,11 @@ describe("Documentation helpers", () => {
...
       z.string().time(),
       z.string().duration(),
-      z.string().cidr(),
-      z.string().ip(),
+      z.string().cidrv4(),
+      z.string().ipv4(),
       z.string().jwt(),
       z.string().base64(),
-      z.string().base64url(),
+      z.string().base64url(), // @todo unclear what happened to it https://github.com/colinhacks/zod/issues/3711
```
RobinTail, line 469: it seems to become `ZodBase64URL`

express-zod-api/src/deep-checks.ts
```diff
@@ -76,9 +88,9 @@ export const hasNestedSchema = (
   if (condition?.(subject)) return true;
   const handler =
     depth < maxDepth
-      ? rules[subject._def[metaSymbol]?.brand as keyof typeof rules] ||
-        ("typeName" in subject._def &&
-          rules[subject._def.typeName as keyof typeof rules])
+      ? rules[
+          globalRegistry.get(subject)?.[metaSymbol]?.brand as keyof typeof rules
```
RobinTail, line 92: ⚠️ this leads to `undefined` string conversion which is an actual schema type in Zod 4

express-zod-api/src/schema-walker.ts
```diff
@@ -49,10 +52,10 @@ export const walkSchema = <
   },
 ): U => {
   const handler =
-    rules[schema._def[metaSymbol]?.brand as keyof typeof rules] ||
-    ("typeName" in schema._def &&
-      rules[schema._def.typeName as keyof typeof rules]);
-  const next = (subject: z.ZodTypeAny) =>
+    rules[
+      globalRegistry.get(schema)?.[metaSymbol]?.brand as keyof typeof rules
```
RobinTail, line 56: ⚠️ this leads to `undefined` string conversion which is an actual schema type in Zod 4

RobinTail:
at this point I fixed all type issues

express-zod-api/src/zts.ts
```diff
@@ -41,48 +58,54 @@ const nodePath = {
...
-    });
-  });
+  const members = Object.entries(def.shape).map<ts.TypeElement>(
+    ([key, value]) => {
+      const isOptional =
+        isResponse && hasCoercion(value)
+          ? value instanceof z.ZodOptional
+          : isResponse
+            ? value._zod.qout
+            : value._zod.qin;
```
RobinTail, line 85: `qin` is not set on ZodOptional.
Filed a fix https://github.com/colinhacks/zod/pull/4153

express-zod-api/src/documentation-helpers.ts
```diff
@@ -235,28 +265,34 @@ const getSupportedType = (value: unknown): SchemaObjectType | undefined => {
...
 ) => {
-  const keys = Object.keys(schema.shape);
-  const isOptionalProp = (prop: z.ZodTypeAny) =>
+  const keys = Object.keys(schema._zod.def.shape);
+  const isOptionalProp = (prop: $ZodType) =>
     isResponse && hasCoercion(prop)
       ? prop instanceof z.ZodOptional
-      : prop.isOptional();
-  const required = keys.filter((key) => !isOptionalProp(schema.shape[key]));
+      : (isResponse ? prop._zod.qout : prop._zod.qin) === "true";
```
RobinTail, line 292: `qin` is not set on ZodOptional.
Filed a fix https://github.com/colinhacks/zod/pull/4153

express-zod-api/src/documentation-helpers.ts
```diff
@@ -498,42 +568,42 @@ const makeNullableType = ({
...
-  schema: z.ZodBranded<z.ZodTypeAny, string | number | symbol>,
-  { next },
-) => next(schema.unwrap());
-
 export const depictLazy: Depicter = (
-  lazy: z.ZodLazy<z.ZodTypeAny>,
+  lazy: $ZodLazy,
   { next, makeRef },
-): ReferenceObject => makeRef(lazy, () => next(lazy.schema));
+): ReferenceObject => makeRef(lazy, () => next(lazy._zod.def.getter()));
```
RobinTail, line 603: this does no longer work, because `getter()` returns new instance each time, so that `WeakSet` does not match them

RobinTail: fixed

RobinTail:
```
      Test Files  10 failed | 34 passed (44)
      Tests  39 failed | 786 passed (825)
```

RobinTail:
```
      Test Files  8 failed | 36 passed (44)
      Tests  33 failed | 792 passed (825)
```

RobinTail:
```
Snapshots  15 failed
Test Files  5 failed | 39 passed (44)
Tests  18 failed | 806 passed (824)
```

RobinTail:
The only issue remains in tests is handling optional props.

RobinTail:
✅ Unit tests done. 
❌ Now there is a build failure.

express-zod-api/src/documentation-helpers.ts
RobinTail: reverted to using `isOptional` so far

RobinTail:
Hooley sheet!

wheee

RobinTail:
`depictString` should also accept `$ZodStringFormat`

express-zod-api/tests/__snapshots__/env.spec.ts.snap
```diff
@@ -0,0 +1,185 @@
...
+`;
+
+exports[`Environment checks > Zod checks/refinements > Snapshot control 'ZodNumber' definition 1`] = `
+{
+  "computed": {
+    "format": "safeint",
+    "maximum": 9007199254740991,
+    "minimum": -9007199254740991,
+    "pattern": /\\^\\\\d\\+\\$/,
+  },
```
RobinTail, line 48: many things for Documentation can be taken from here

RobinTail: this would be a feature, not migration of the current logic

RobinTail:
Ok. At this point the blocker is the third party issue on refinements missing in shorthand schemas.
Could also improve coverage while it's not fixed.

express-zod-api/src/metadata.ts
```diff
@@ -1,38 +1,37 @@
...
-    ([destExample, srcExample]) =>
-      typeof destExample === "object" && typeof srcExample === "object"
-        ? R.mergeDeepRight({ ...destExample }, { ...srcExample })
-        : srcExample, // not supposed to be called on non-object schemas
-  );
-  return result;
+  const srcMeta = src.meta()?.[metaSymbol];
+  const destMeta = dest.meta()?.[metaSymbol];
+  if (!srcMeta) return dest; // ensure metadata in src below
+
```
RobinTail, line 22:
```suggestion
```

express-zod-api/src/metadata.ts
```diff
@@ -1,38 +1,37 @@
-import { combinations } from "./common-helpers";
-import { z } from "zod";
 import * as R from "ramda";
+import { z } from "zod";
+import { combinations } from "./common-helpers";
```
RobinTail, line 3: reduce diff

```suggestion
import { combinations } from "./common-helpers";
import { z } from "zod";
import * as R from "ramda";
```

express-zod-api/src/zod-plugin.ts
```diff
@@ -8,79 +8,93 @@
...
-const exampleSetter = function (
-  this: z.ZodType,
-  value: (typeof this)["_input"],
-) {
-  const copy = cloneSchema(this);
-  copy._def[metaSymbol]!.examples.push(value);
-  return copy;
+const exampleSetter = function (this: z.ZodType, value: z.input<typeof this>) {
+  const { examples, ...rest } = this.meta()?.[metaSymbol] || { examples: [] };
+  examples.push(value);
```
RobinTail, line 58: check if it does not push to registry here

express-zod-api/src/zts.ts
```diff
@@ -41,48 +58,54 @@ const nodePath = {
   optional: R.path(["questionToken" satisfies keyof ts.TypeElement]),
 };
 
-const onLiteral: Producer = ({ value }: z.ZodLiteral<LiteralType>) =>
-  makeLiteralType(value);
+const onLiteral: Producer = ({ _zod: { def } }: $ZodLiteral) =>
+  f.createUnionTypeNode(
```
RobinTail, line 62: maybe should be an early exit for the case of a single value

RobinTail (commented):
okay

PR #2539, opened 2025-04-13
Requiring `express@^5.1.0`

RobinTail:
That version was the first on v5 marked as `latest`.
So I believe it would be better to focus on it instead of `5.0.1`


RobinTail (commented):
ok

PR #2547, opened 2025-04-17
Delegating depiction to Zod 4

RobinTail:
Exploring the possibility to delegate OpenAPI compatible depiction to JSON Schema based method of Zod 4.

Current issues:
- Can not depict `BigInt` with throwing error if it's within `ZodLiteral`
  - https://github.com/colinhacks/zod/issues/4220 
- Numeric limits are wrong:
  - exclusive, while should be inclusive
  - missing at all (`z.number().int().positive()`)
  - https://github.com/colinhacks/zod/pull/4074#discussion_r2048393003
  - suggested fix https://github.com/colinhacks/zod/pull/4224
- Overrides act after default depiction, not before, so that much has to be rewritten

PR #2553, opened 2025-04-20
Helper to ensure valid `type`

RobinTail:
Addressing todo from #2547 
and improving `makeSupportedType`


RobinTail (commented):
this can also be backported to v23

PR #2556, opened 2025-04-21
Tests for overriders and minor adjustments

RobinTail:
Instead of testing Zod's depiction, transforming current tests into the ones testing actions of overriders individually.


RobinTail (commented):
ok

PR #2557, opened 2025-04-21
Removing overrides for enums and literals

RobinTail:
`type` is not needed when you have `const` or `enum` entries.
So this one can be fully delegated to Zod 4.


RobinTail (commented):
ok

PR #2562, opened 2025-04-21
Exp: native storage for input examples

RobinTail:
New metadata system of Zod 4 finally supports examples.
However, those examples are typed as `z.output<T>`, while the proprietary ones used to be `z.input<T>`, so that the could be transformed when necessary or taken as is depending on the case.

this requires investigation.

### DIFF
--- a/example/example.documentation.yaml
+++ b/example/example.documentation.yaml
@@ -107,10 +107,10 @@ paths:
           required: true
           description: PATCH /v1/user/:id Parameter
           schema:
-            type: string
-            minLength: 1
             examples:
               - "1234567890"
+            type: string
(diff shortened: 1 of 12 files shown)


RobinTail:
switching to output would affect `getExamples()` method

RobinTail:
maybe... just MAYBE
the plugin could transform examples when `.tranform()` method applied and set them to the resulting schema.

RobinTail:
first need to separate registries, then will decide to move it or not

RobinTail:
the transformation examples fixed in Zod, but will start this PR over anyway due to much conflicts

PR #2563, opened 2025-04-22
Proprietary metadata registry

RobinTail:
Instead of using the global registry for internal purposes under a symbol, I wanna move those things into proprietary one.

PR #2566, opened 2025-04-22
Exp: Split registries

RobinTail:
Instead of #2563 
But there is an issue.
You can't do `.example().describe()` and then retrieve the examples of it.


RobinTail:
this requires more complicated side effects

RobinTail:
Actually, it may be just enough to move anchor from `check` to `clone`

RobinTail:
attaching to clone seems to be messy

PR #2574, opened 2025-04-24
Using `@zod/core` types

RobinTail:


RobinTail:
I should postpone it for later and come up with a clear definition on where it should and should not be applied.
For now it seems to be everything related to traversing schemas.

PR #2575, opened 2025-04-24
Removing arguments from schema types

RobinTail:
Instead of #2574 (partially)
Due to defaults implemented by Zod 4


RobinTail (commented):
ok

PR #2581, opened 2025-04-25
Implementing `unref()` helper

RobinTail:
https://github.com/colinhacks/zod/issues/4275
Extracting that into a helper.

### DIFF
--- a/example/example.documentation.yaml
+++ b/example/example.documentation.yaml
@@ -518,6 +518,8 @@ paths:
           application/octet-stream:
             schema:
               type: string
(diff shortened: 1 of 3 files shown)


example/example.documentation.yaml
```diff
@@ -518,6 +518,8 @@ paths:
           application/octet-stream:
             schema:
               type: string
+              required:
+                - raw
```
RobinTail, line 522: should not be here

RobinTail (commented):
ok

PR #2604, opened 2025-05-05
v24: Utilize `optionality`

RobinTail:
All properties are now required in Zod 4 unless they are `.optional()` or `.default()`.
That fact is now reflected by `._zod.optionality`
Direction matters.

External bugs:
- https://github.com/colinhacks/zod/issues/4322

PR #2607, opened 2025-05-06
Optionals: Restore `undefined`, but depending on `optionality`

RobinTail:
Partially reverts 773b212f51921da7704105c8a45cf2373908e752 from #2604 
But this implementation does it depending on `optionality` which in its turn depends on direction.
This takes Collin explanaition into account https://x.com/colinhacks/status/1919292504861491252
And adds env test to ensure that.

### DIFF
--- a/express-zod-api/tests/env.spec.ts
+++ b/express-zod-api/tests/env.spec.ts
@@ -1,6 +1,5 @@
 import createHttpError from "http-errors";
 import * as R from "ramda";
-import { expectTypeOf } from "vitest";
 import { z } from "zod";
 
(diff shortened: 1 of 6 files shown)


express-zod-api/tests/env.spec.ts
```diff
@@ -1,5 +1,6 @@
 import createHttpError from "http-errors";
 import * as R from "ramda";
+import { expectTypeOf } from "vitest";
```
RobinTail, line 3:
```suggestion
```

RobinTail (commented):
ok

PR #2622, opened 2025-05-12
Using `undici` 7

RobinTail:


RobinTail:
it seems work faster for graceful shutdown test

example/package.json
```diff
@@ -18,6 +18,6 @@
   },
   "devDependencies": {
     "@types/swagger-ui-express": "^4.1.8",
-    "undici": "^6.19.8"
+    "undici": "^7.10.0"
```
RobinTail: revisit todo in index.spec.ts

PR #2546, opened 2025-04-17
v24 is for Ashley Burton

RobinTail:
Related to #2550

## Dedication

image

Ashley Burton was a 37 years young black transgender woman, a hairstylist and a makeup artist in Atlanta. 
She was shot to death at her apartment complex April 11th 2023. 
Her body was found shortly after 4:30 a.m. in the breezeway of her building.

> She ran out of the house, hollering and screaming, beating on doors.
> This got to be personal. You shot her in her house, then you followed her outside and shot her.

— _her cousin Ivory Carter told Atlanta TV station WXIA_.

Burton and Carter grew up together in South Carolina.

> She was a sweetheart, she wasn’t nasty. She wasn’t disrespectful. She just wanted to live her life.
> I’m tired of all these incidents with transgender women just being pushed up under the rug. We are human beings.

— _Carter said_.

Patrick Burton, the victim’s brother, told Atlanta station WAGA that “Ashley was very loved all the way across the board, like from South Carolina to Atlanta.”

Darius Mills, 31, was arrested in June in connection with the crime. He faces charges of murder, armed robbery, possession of a firearm by a convicted felon, and possession of a firearm during the commission of a felony.

https://www.advocate.com/crime/ashley-burton-trans-woman-death

Transgender women suffer [too frequently](https://www.congress.gov/118/meeting/house/117016/documents/HMKP-118-JU00-20240321-SD015.pdf) from transphobic violence and cruelty, being the less protected social group. I'd like to raise an awareness of this problem. Humans should be creators — not killers. But most importantly, I want every transgender girl to have an opportunity to create applications quickly and, in general, learn to write code easily in order to receive job offers and leave dangerously transphobic territories for more favorable and civilized ones, and live happily there. Protect transgender women.

## Modern impact

image

The Trump administration's policies have significantly impacted transgender individuals, particularly those who are most vulnerable, by dismantling protections and fostering environments that increase the risk of discrimination and violence. Here's an overview of how these policies have affected the transgender community:

### 1. Executive Order 14168: Erasure of Transgender Recognition

[...]

PR #2625, opened 2025-05-15
Testing `zod` 3.25

RobinTail:
ok, well...
https://x.com/colinhacks/status/1922341985752646129

So, it's suggested to refactor everything to support both v3 and v4 of Zod, as temporary solution for migrating to v4.

I'm not sure about that. They are so different entirely, so that it would require some sort of provider inbetween, but since all of that it temporary, I don't see much reason for such an effort.

another problem is proprietary schemas (`ez`), that should either also be duplicated, or versionized, otherwise it would cause a mix of versions within composite schemas.

So, overall it's a mess.

Related to #2546


yarn.lock
```diff
@@ -3123,7 +3123,7 @@ yocto-queue@^0.1.0:
   resolved "https://registry.yarnpkg.com/yocto-queue/-/yocto-queue-0.1.0.tgz#0294eb3dee05028d31ee1a5fa2c556a6aaf10a1b"
   integrity sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==
 
-zod@^3.24.4:
-  version "3.24.4"
-  resolved "https://registry.yarnpkg.com/zod/-/zod-3.24.4.tgz#e2e2cca5faaa012d76e527d0d36622e0a90c315f"
-  integrity sha512-OdqJE9UDRPwWsrHjLN2F8bPxvwJBK22EHLWtanu0LSYr5YqzsaaW3RMgmjwr8Rypg5k+meEJdSPXJZXE/yqOMg==
+zod@^3.25.0-beta.20250517T081109:
+  version "3.25.0-canary.20250504T221648"
```
RobinTail, line 3127: this one is wrong, should be beta

PR #2631, opened 2025-05-17
Using v4 from 3.25

RobinTail:
The title of this PR is insane :)

### DIFF
--- a/express-zod-api/tests/deep-checks.spec.ts
+++ b/express-zod-api/tests/deep-checks.spec.ts
@@ -45,7 +45,7 @@ describe("Checks", () => {
       ).toBeUndefined();
     });
 
-    test("should finish early bottom to top", () => {
+    test("should finish early (from bottom to top)", () => {
       const subject = z.object({
         one: z.object({
           two: z.object({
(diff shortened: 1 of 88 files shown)


RobinTail:
it seems that because it's changed to top-to-bottom, overrides of intersection no longer get the depiction inside `allOf`

RobinTail:
It's getting worse and worse with every update.
1 bug fixes, 3 new added

RobinTail:
since the order changed, things like `z.literal().nullable()` now get overriden as ZodNullable ahead of ZodLiteral, so there is no way to set `type` of the literal before making it nullable, like we used to do before.

RobinTail:
I was told that he fixed "overrides". Let's see.

express-zod-api/tests/deep-checks.spec.ts
```diff
@@ -45,7 +45,7 @@ describe("Checks", () => {
       ).toBeUndefined();
     });
 
-    test("should finish early (from bottom to top)", () => {
+    test("should finish early bottom to top", () => {
```
RobinTail, line 48:
```suggestion
    test("should finish early (from bottom to top)", () => {
```

RobinTail (commented):
OK, wow
can't believe we fixed it

PR #2633, opened 2025-05-19
Changes to optionality

RobinTail:
- address replaced `optionaility` with `optin` and `optout`
- that also fixed question marks on optional schemas with transformations
  - https://github.com/colinhacks/zod/issues/4322
- new bug: https://github.com/colinhacks/zod/pull/4407
- also, it appears to change the way `.meta()` works: now it seems to merge the metadata


RobinTail (commented):
ok

PR #2637, opened 2025-05-19
Branding: no altering refinements

RobinTail:
For that, brand should not be a part of metadata, but somewhere else.
Presumably, in `def` because it withstands `.clone()`.
Or in `_zod.bag` as suggested by Colin.

### DIFF
--- a/express-zod-api/src/metadata.ts
+++ b/express-zod-api/src/metadata.ts
@@ -35,9 +35,7 @@ export const mixExamples = <A extends z.ZodType, B extends z.ZodType>(
   });
 };
 
-export const getBrand = (
-  subject: $ZodType,
-): string | number | symbol | undefined => {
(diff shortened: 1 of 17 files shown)


RobinTail:
experiment is a success

express-zod-api/src/metadata.ts
```diff
@@ -34,3 +34,16 @@ export const mixExamples = <A extends z.ZodType, B extends z.ZodType>(
     [metaSymbol]: { ...destMeta?.[metaSymbol], examples },
   });
 };
+
+export const getBrand = (
```
RobinTail, line 38: perhaps it should rather go to common helpers

RobinTail: let's keep it for no

RobinTail (commented):
good

PR #2632, opened 2025-05-18
Using native storage for examples

RobinTail:
New approach to #2562.

I'm considering to remove the feature of transforming examples and devote to users its proper placement.

The native metadata examples are typed as `z.output<>` of the schema, which is the opposite of what it used to be established by the framework.

PR #2643, opened 2025-05-20
Using native storage for default label

RobinTail:
While working on #2632 I realized that native storage is actually the right place for it


RobinTail (commented):
good!

PR #2644, opened 2025-05-20
Ref: no reassigning meta

RobinTail:
The stable Zod 4 merges metadata, not replaces it.
So it's now safe to set only one entry without spreading the prev state.


RobinTail (commented):
ok

PR #2649, opened 2025-05-21
feat(v24): metadata argument for `ez.dateIn()` and `ez.dateOut()`

RobinTail:
This should bypass the need for parsing examples OR using `$brand` to set them

### DIFF
--- a/example/example.documentation.yaml
+++ b/example/example.documentation.yaml
@@ -153,7 +153,7 @@ paths:
                   externalDocs:
                     url: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString
                   examples:
-                    - 1963-04-21T00:00:00.000Z
+                    - 1963-04-21
               required:
                 - key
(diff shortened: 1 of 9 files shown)


example/example.documentation.yaml
```diff
@@ -163,7 +163,6 @@ paths:
                 value:
                   key: 1234-5678-90
                   name: John Doe
-                  birthday: 1963-04-21T00:00:00.000Z
```
RobinTail, line 166: check why it's gone

RobinTail: because pulling examples in mixer and that does not involve transformations.
while instead should probably do it in `flattenIO()` and then revert c1e111c612773703551d4ede65944d8596e62368

RobinTail: addressed in #2651

RobinTail (commented):
good.
need to update readme and changelog

RobinTail (commented):
ok

PR #2651, opened 2025-05-22
Ref: Move mixing and pulling request examples to postprocessing

RobinTail:
The need of it discovered during #2649 

Examples set before transformation are not getting to body depiction.
see https://github.com/RobinTail/express-zod-api/pull/2649#discussion_r2101460393

this is because mixing and pulling before in preprocess, while to should be devoted to Zod to depict them first, and then do mixing/pulling in a postprocess, e.g. `flattenIO`


RobinTail (commented):
fantastic

PR #2655, opened 2025-05-23
Feat: `beforeExit`

RobinTail:
based on #2650


HeikoOsigus:
Wow, i am really impressed how fast this was done!!!
That would perfectly suit my needs.

PR #2660, opened 2025-05-24
`createServer()`: Warn when no servers configured

RobinTail:
Missed that one.
Non-breaking version


RobinTail:
should add a test

PR #2661, opened 2025-05-24
Natively prettified ZodErrors in response

RobinTail:


RobinTail:
these icons in native prettifying are too opinionated.
i'd rathe change my own one.

PR #2663, opened 2025-05-25
feat(v24): Error tree

RobinTail:
https://zod.dev/error-formatting#ztreeifyerror


example/example.documentation.yaml
```diff
@@ -751,6 +761,106 @@ components:
...
+        properties:
+          type: object
+          propertyNames:
+            type: string
+          additionalProperties:
+            $ref: "#/components/schemas/Schema2"
+      required:
+        - errors
+      additionalProperties: false
+    Schema3:
```
RobinTail, line 784: perhaps externals `defs` storage could fix it

RobinTail:
should not overload the default RH

PR #2665, opened 2025-05-26
Adjust error message format in response

RobinTail:
Switching to zod core's `toDotPath()` in `getMessageFromError()`.

I'm not using `z.prettifyError()` because its prettiness to me is questionable due to:
- icons/symbols
- new lines
- offsets 

In my opinion such formatting should be delegated to UI.
For that I'm planning #2663


RobinTail (commented):
ok

PR #2667, opened 2025-05-26
Renaming `ez.file()` to `ez.download()`

RobinTail:
Here I'm addressing the possible confusion between `ez.file()` and the new schema of Zod 4 `z.file()`.

The proprietary one is designed for depicting file content in response having the form of string or a buffer.
For the sake of consistency with `ez.upload()` I'm renaming it to `ez.download()`.


RobinTail (commented):
perhaps, it would be even better to remove `ez.file()` completely, but expose the buffer schema with a suggestion to use `.meta()` for setting customized props depending on particular needs.

PR #2705, opened 2025-06-04
Simpler type for `ZodObject::remap()`

RobinTail:
Can be simplified a little more for Zod 4

## Summary by CodeRabbit

- **Refactor**
  - Simplified internal type annotations for the remap method, resulting in cleaner and more general type signatures. No changes to visible functionality or user experience.


RobinTail (commented):
neat

PR #2759, opened 2025-06-18
Exp: Using `pnpm` alias approach instead of min catalog

RobinTail:
I'm trying to exclude `compat-test` workspace from renovate

## Summary by CodeRabbit

- **Chores**
  - Updated dependency versions in the compatibility test package to use explicit versions.
  - Removed the minimum version catalog from workspace configuration.
  - Simplified Renovate configuration by excluding the compatibility test directory from automated dependency updates.


RobinTail (commented):
let's try it

PR #2761, opened 2025-06-19
Exp: Improving final intersection 

### LINKED ISSUE #2760: Accessing undefined input properties returns never type instead of compilation error
# Description

When accessing undefined properties on the `input` object results in a `never` type instead of a TypeScript compilation error. 

```ts
const helloWorldEndpoint = defaultEndpointsFactory.build({
  input: z.object({
    foo: z.string(),
  }),
  output: z.object({
    foo: z.string(),
  }),
  handler: async ({ input: { bar, foo } }) => {
    console.log(bar) // <----------------------- bar: never
    return { foo };
  },
});
```

## Preferred Behavior

I'd prefer TypeScript to throw a clear compilation error such as:
```
Property 'foo' does not exist on type ...'
```

This would make debugging easier and catch typos at compile time rather than having to discover the `never` type through hovering or other means.

The `never` type essentially silences what should be a helpful TypeScript error, making it harder to catch mistakes during development.

## Context

- `express-zod-api` version: 24.4.1

RobinTail:
Fixes #2760

### DIFF
--- a/express-zod-api/src/endpoints-factory.ts
+++ b/express-zod-api/src/endpoints-factory.ts
@@ -31,7 +31,11 @@ interface BuildProps<
   /** @desc The schema by which the returns of the Endpoint handler is validated */
   output: OUT;
   /** @desc The Endpoint handler receiving the validated inputs, returns of added Middlewares (options) and a logger */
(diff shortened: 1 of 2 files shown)


RobinTail:
In favor of #2762

PR #2763, opened 2025-06-22
Using `vitest` 4

RobinTail:


RobinTail:
something affects `@types/node` resolution in migration workspace

RobinTail:
something isn't right here, maybe should postpone to v25

PR #2786, opened 2025-07-02
Chores on automated deps upgrade

RobinTail:
## Summary by CodeRabbit

* **Chores**
  * Updated dependency management workflow to upgrade all dependencies except those in the compat-test workspace.
  * Adjusted package version overrides to manage specific dependencies more effectively.


package.json
```diff
@@ -36,7 +36,8 @@
     "vitest": "^3.2.3"
   },
   "resolutions": {
-    "@scarf/scarf": "npm:empty-npm-package@1.0.0"
+    "@scarf/scarf": "npm:empty-npm-package@1.0.0",
+    "vite": "^6"
```
RobinTail: this should be reverted in #2741

PR #2815, opened 2025-07-13
Using `vitest` 4

RobinTail:
Instead of #2763 

It became insanely fast!

- changes to the pool options (beta 19): https://github.com/AriPerkkio/vitest/blob/713b9f40c3fbd42ce97a28297c64668fad06fd9f/docs/advanced/pool.md


RobinTail:
The number of removed dependencies is fantastic

PR #2829, opened 2025-07-18
Fix #2828 for v23

### LINKED ISSUE #2824: `.use()` on endpoints factory types not correct?
# Description

Been upgrading through several versions of this library recently, from what I can tell this issue started in `23.0.0` but is still an issue in the latest version.

When trying to attach express middleware to a specific route, in this case it is `express-rate-limit`, I get a type error, for this example it's:

```
Argument of type 'RateLimitRequestHandler' is not assignable to parameter of type '(request: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, response: Response<any, Record<string, any>, number>, next: NextFunction) => void | Promise<...>'.
  Type 'unknown' is not assignable to type 'void | Promise<void>'.ts(2345)
```

I briefly tried out the library in the documentation example too (`express-oauth2-jwt-bearer`) and had the same error there, so I assume it's not library specific. At first I assumed it might've been the library not being compatible with express v5 or similar, but the types work fine if I attach the middleware inside of the `beforeRouting` hook.

## Expected

No type errors and the code to build.

## Reproduction

```
import rateLimit from 'express-rate-limit';
const limiter = rateLimit();
const example = generic.endpointsFactory() // returns my EndpointsFactory instance
  .use(limiter) // <-- type error happens here
  .build({
    method: 'get',
    input: z.object({}),
    output: z.object({}),
    handler: async () => null,
  });
```
## Context

- Node.js version: 22.15.0
- `express-zod-api` version: 24.6.0

RobinTail:
Fixes #2824 for v23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Version 23
 
+### v23.6.3
+
+- Compatibility fix for recently changed type of Express native middleware:
+  - Fixes error `TS2345` when passing over an Express middleware to:
+    - `EndpointsFactory::use()`;
+    - `EndpointsFactory::addExpressMiddleware()`;
+  - The issue caused by `@types/express-serve-static-core` v5.0.7;
+  - The issue reported by [@zoton2](https://github.com/zoton2).
+
 ### v23.6.2
 
 - Fixed a bug about missing CORS headers in case of request parser errors:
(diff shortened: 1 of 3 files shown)


RobinTail (commented):
ok

PR #2844, opened 2025-07-26
Accessing core types via `z.core`

RobinTail:
Previously, I separated core types imports because they were intended to be published in a separate package `@zod/core`.
However, since stable v4 rolled out, that idea was abandoned, so that it's now safe to refer to it using the `z.core` without additional imports.
This was suggested by @coderabbitai 

## Summary by CodeRabbit

* **Refactor**
  * Updated internal type references to use the public `z.core` namespace instead of direct imports from internal Zod paths. This affects type annotations and function signatures throughout the codebase, including helpers, schemas, documentation utilities, and tests.

* **Tests**
  * Adjusted test files to use the new `z.core` type references, ensuring consistency with the updated type import approach. 

No changes to application logic or user-facing features.

### DIFF
--- a/express-zod-api/src/common-helpers.ts
+++ b/express-zod-api/src/common-helpers.ts
@@ -1,7 +1,6 @@
 import { Request } from "express";
 import * as R from "ramda";
 import { z } from "zod";
-import type { $ZodTransform, $ZodType } from "zod/v4/core";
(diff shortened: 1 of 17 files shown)


RobinTail:
@coderabbitai review

PR #2857, opened 2025-07-29
Fix negative response MIME type for `arrayResultHandler`

RobinTail:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the negative response MIME type for deprecated error handling to use "text/plain" instead of "application/json" in both documentation and response handling.

* **Documentation**
  * Updated the changelog to reflect the fix for the negative response MIME type.
  * Adjusted API documentation to specify "text/plain" as the content type for certain error responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Version 24
 
+### v24.7.2
+
+- Fixed the negative response MIME type for ~~`arrayResultHandler`~~ (deprecated entity):
+  - Should have been `text/plain`.
+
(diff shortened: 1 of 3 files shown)


RobinTail:
@coderabbitai

> 458-465: HEAD responses generally shouldn’t declare a body

No body for a successfully handled request (status code 200).
But if there was an error while handling the request (status code 400), then it can be body of the response to explain the error. Isn't it?

RobinTail:
Ok. Thanx. I will address it in another PR then, @coderabbitai

PR #2859, opened 2025-07-29
Removing body depiction from negative response to `HEAD` requests

### LINKED ISSUE #2858: Negative response to HEAD request should also have no body
- HEAD responses must return the **same headers** that a GET request would return (including `Content-Type`, `Content-Length`, etc.)
- But the response body must always be **empty**, even for error responses
- This allows clients to get metadata about the resource without transferring the actual content

So for a 400 error:
- **GET** would return: headers + error message body
- **HEAD** would return: same headers (indicating there would be an error message) + empty body

The headers themselves (like `Content-Type: text/plain`) communicate that there would be an error message, but the actual error text shouldn't be included in the HEAD response body.

This behavior is consistent across all HTTP status codes - HEAD responses should always have empty bodies, but the headers should reflect what a corresponding GET request would return.

_Originally posted by @coderabbitai[bot] in https://github.com/RobinTail/express-zod-api/issues/2857#issuecomment-3133952165_

RobinTail:
Fixes #2858 

Noticed by @coderabbitai in https://github.com/RobinTail/express-zod-api/pull/2857#issuecomment-3133952165

### DIFF
--- a/express-zod-api/src/common-helpers.ts
+++ b/express-zod-api/src/common-helpers.ts
@@ -11,7 +11,6 @@ import {
   Method,
   CORSMethod,
 } from "./method";
-import { ResponseVariant } from "./api-response";
 
 /** @desc this type does not allow props assignment, but it works for reading them when merged with another interface */
 export type EmptyObject = z.output<EmptySchema>;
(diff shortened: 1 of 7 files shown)


express-zod-api/src/common-helpers.ts
```diff
@@ -143,7 +142,4 @@ export const isProduction = R.memoizeWith(
   () => process.env.NODE_ENV === "production", // eslint-disable-line no-restricted-syntax -- memoized
 );
 
-export const doesImplyContent = (
-  method: ClientMethod,
-  responseVariant: ResponseVariant,
-) => !(method === "head" && responseVariant === "positive");
+export const doesImplyContent = (method: ClientMethod) => method !== "head";
```
RobinTail: perhaps it could be inline

RobinTail: or it could also verify MIME types

RobinTail (commented):
nice

PR #2860, opened 2025-07-29
Reusing `R.Remap` type

RobinTail:
I suggested this type to Ramda maintainers for `R.renameKeys()` method

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal type handling for object remapping, reducing dependencies and streamlining type definitions.
* **Chores**
  * Removed unused utility types to improve maintainability.


express-zod-api/src/zod-plugin.ts
```diff
@@ -68,7 +69,7 @@ declare module "zod/v4" {
       mapping: U,
     ): z.ZodPipe<
       z.ZodPipe<this, z.ZodTransform>, // internal type simplified
-      z.ZodObject<Remap<Shape, U, V> & Intact<Shape, U>, Config>
+      z.ZodObject<R.Remap<Shape, U> & Intact<Shape, U>, Config>
```
RobinTail, line 72: check if it's actually exposed/resolved into the build

RobinTail: it's not. but there is an option in `tsup`: `dts.resolve`, however, it's not documented how exactly it works (it does not)

RobinTail:
ok, this might require to include `@types/ramda` into the list of production dependencies, which I don't like

RobinTail:
I think I found the reason:
> --dts-resolve will ignore packages specified in the dependencies field in package.json, meaning they're always externalized

https://dev.to/egoist/rollup-dts-file-using-tsup-2579#fn1

RobinTail:
Perhaps it would be better to get back to it if I decide to detach the plugin.

PR #2890, opened 2025-08-16
Root `tsconfig` and ESLint config in typescript

RobinTail:


RobinTail:
eslint config file in compat test will remain JS, because TS-based one was added in 9.24.0

RobinTail (commented):
should be fine

PR #2912, opened 2025-08-31
Add missing Node engine restrictions to `zod-plugin`

RobinTail:
## Summary by CodeRabbit

- Documentation
  - Added changelog entry for v2.0.0 documenting supported Node.js versions: ^20.19.0 || ^22.12.0 || ^24.0.0.

- Chores
  - Declared supported Node.js versions in package metadata to improve install-time compatibility checks; users on unsupported Node versions may see warnings or install blocks.
  - Bumped package version to 2.0.0-beta.2; no functional or API changes.


RobinTail:
My concern here is that technically it is a breaking change.

RobinTail:
✅ checked the published code

PR #2920, opened 2025-09-06
Exp: RH constraints

RobinTail:
Exploring a possibility to establish constraints between `Endpoint::output` and `ResultHandler::positive`


RobinTail:
overcomplicates `::buildVoid()`

PR #2954, opened 2025-09-21
v24: Replacing usage of `example()` method in framework sources

RobinTail:
A user reported that they're having troubles with module augmentation in their environment, but they aren't even using the `example()` method. It's found to be used in the framework itself and this PR should fix it, so that the framework would be usable for them even without the plugin features.

### DIFF
--- a/express-zod-api/src/result-handler.ts
+++ b/express-zod-api/src/result-handler.ts
@@ -112,15 +112,17 @@ export const defaultResultHandler = new ResultHandler({
     }
     return responseSchema;
   },
-  negative: z
-    .object({
+  negative: () => {
+    const schema = z.object({
       status: z.literal("error"),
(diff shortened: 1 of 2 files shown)


express-zod-api/src/result-handler.ts
```diff
@@ -112,15 +112,17 @@ export const defaultResultHandler = new ResultHandler({
     }
     return responseSchema;
   },
-  negative: z
-    .object({
+  negative: () => {
```
RobinTail: This is static, but it makes a new instance in the registry each time. I think it could cause memory leak.

RobinTail: fixed by c50e58f3

RobinTail:
@coderabbitai review

RobinTail:
✅ QA passed

PR #2973, opened 2025-09-28
rm `strictNullChecks` globally

RobinTail:
because already covered by `strict` by `@tsconfig/node20` base

## Summary by CodeRabbit

* Chores
  * Updated TypeScript configuration by disabling strict null checks, reducing nullability strictness during development and builds.
  * Tooling-only adjustment to improve developer experience and compatibility; no changes to public APIs.
  * No user-facing impact: application behavior, features, performance, and UI remain unchanged.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "noImplicitAny": true,
     "noImplicitOverride": true,
+    "noUncheckedSideEffectImports": true,
     "types": ["vitest/globals"]
   }
 }


RobinTail:
> thereby relaxing how TypeScript enforces null and undefined handling

@coderabbitai, no relaxing. The `tsconfig.json` in root of the project extends from `@tsconfig/node20/tsconfig.json` that contains `strict: true`, that implies `strictNullChecks` too. So it's just a cleanup.

PR #2982, opened 2025-10-02
fix(zod-plugin): Support Zod's CJS

### LINKED ISSUE #2981: v25: example is not a function
# Description

Upgrading from v24 to v25 appears to break examples in generated documentation. I created a git repo which reproduces this issue: https://github.com/squishykid/express-zod-api-example-issue

There are two ways this library allows developers to provide examples:
- `.example()`
- `.meta()`

Upon upgrading from 24.7.3 to 25.0.0 it appears that both of these methods of providing examples break in different ways.

Examples supplied via the .meta() method simply vanish silently from the generated documentation. Perhaps zod's `globalRegistry` is not as global any more? I notice that examples (default error messages) added by this library continue to be included in the documentation yaml output.

Examples supplied via the `.example()` method cause a compile error after upgrading from v24.7.3 to v25.0.0:
```
}).example({ name: 'bingo' });
   ^

TypeError: zod_1.z.object(...).example is not a function
```

## Expected

I would have expeted `.example()` and `.meta()` to both produce examples in the documentation output, regardless of using v24.x or v25.x of this library.

## Reproduction

I have created a repository here which extends a previous minimal example for zod bugs.

The repo currently uses v24.7.3, but upon upgrading to v25.0.0 the bug will be triggered.

https://github.com/squishykid/express-zod-api-example-issue

## Context

- Node.js version: 22.20.0
- `express-zod-api` version: v24.7.3 to v25.0.0

RobinTail:
fixes #2981

### DIFF
--- a/zod-plugin/tsdown.config.ts
+++ b/zod-plugin/tsdown.config.ts
@@ -4,7 +4,7 @@ export default defineConfig([
   {
     entry: "src/index.ts",
     minify: true,
-    attw: { profile: "esmOnly", level: "error" },
+    attw: { level: "error" },
     banner: {
       dts: "import './augmentation.js';",
     },


RobinTail:
🚀 Published `2.1.0-beta.1` of the plugin, can be QAed by using `overrides` or `resolutions` in package manager.

RobinTail:
❌ does not fix the issue directly because `express-zod-api` still `import`s the plugin
☑️ however, when installing on consumer side and importing the plugin BEFORE the framework, it fixes the issue

RobinTail:
not enough

PR #2990, opened 2025-10-04
Exp: `erasableSyntaxOnly`

RobinTail:


RobinTail:
weird error matching problem in Node 20

RobinTail:
somehow that problem depends on `--coverage` flag 😮 
and it even persists in vitest 4: #2815

RobinTail:
Doing #3055 instead

PR #3049, opened 2025-11-08
feat(v26): method only routes

RobinTail:
Due to #2938 

This will set us free from the requirement to write slash to the keys when we imply method depending routing.

Breaking change: `Integration` requires `config` parameter.

## Summary by CodeRabbit

* **New Features**
  * Added `methodLikeRouteBehavior` configuration option to control interpretation of method-like routing keys ("method" or "path")
  * Added `hasHeadMethod` option for endpoint documentation configuration

* **Refactor**
  * Integration constructor now requires a `config` parameter during initialization
  * Routing syntax simplified to use cleaner method-based keys


RobinTail:
@coderabbitai review

PR #3056, opened 2025-11-10
feat: exposing `BuildProps` and `Endpoint` type

mxsdev:
We could use these to write custom wrappers around `build`.

Feel free to modify this PR before merging etc, thanks!

## Summary by CodeRabbit

* **New Features**
  * Exported `BuildProps` type for public use, allowing developers to access endpoint builder configuration types.
  * Exported `Endpoint` type for public use, enabling developers to reference endpoint type definitions directly.

### DIFF
--- a/express-zod-api/src/endpoints-factory.ts
+++ b/express-zod-api/src/endpoints-factory.ts
@@ -27,7 +27,7 @@ import {
   defaultResultHandler,
 } from "./result-handler.ts";
 
-interface BuildProps<
+export interface BuildProps<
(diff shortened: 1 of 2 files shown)


RobinTail:
> custom wrappers around build

what could be that customization, @mxsdev ?

RobinTail:
- Endpoint instance is not supposed to be assigned with type arguments, instead they should be inferred from the factory;
- the problem to be addressed by this PR is unclear;
- exposing those entities would require to maintain their changes according to SemVer;

PR #3090, opened 2025-11-24
Run CJS test without `tsx`

RobinTail:
- precompile using TSC to ensure `require()` in executed code
- because TSX does not run real CJS, it transforms the code before running
- this should improve testing and make the framework tested better in CJS env
- the tsconfig of CJS test is now detached from the root one

## Release Notes

* **Tests**
  - Enhanced test infrastructure with TypeScript compilation as a pre-test step, followed by automatic cleanup
  - Updated test execution to run against compiled JavaScript instead of TypeScript files directly
  - Added TypeScript configuration for the test environment

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/cjs-test/tsconfig.json
+++ b/cjs-test/tsconfig.json
@@ -1,6 +1,4 @@
 {
   "extends": "@tsconfig/node20/tsconfig.json",
-  "compilerOptions": {
-  },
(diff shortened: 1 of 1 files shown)


RobinTail:
@coderabbitai review

PR #3125, opened 2025-12-11
feat: Custom Diagnostics

RobinTail:
Related to #3107 

This is for dev mode only.
A custom hook to run on startup for every route to perform user customized verifications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Version 26
 
+### v26.3.0
+
+- New configuration hook `routeCheck` for implementing custom diagnostics of your routing:
+  - The assigned function is called during the API startup in development mode only (not in production);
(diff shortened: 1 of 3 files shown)


express-zod-api/src/diagnostics.ts
```diff
@@ -94,5 +97,6 @@ export class Diagnostics {
     }
     this.#checkSchema(ref, endpoint, { method, path });
     this.#checkPathParams(ref, endpoint, path, { method });
+    this.customCheck?.(method, path, endpoint);
```
RobinTail: I believe they would also need a `logger` there

RobinTail: perhaps an object-accepting function would be also more suitable for maintenance

RobinTail: added in 9d3a9128

RobinTail:
not necessary

PR #3139, opened 2025-12-16
Memory usage optimization

RobinTail:
Trying to import `typescript` dynamically in order to avoid excessive memory consumption.
Addresses #3126 

Using `createRequire` for sync approach — This is kinda a workaround, because the proper implementation would need:
- `import()` and `async`, which is breaking
- or moving to another entrypoint or another package, which is breaking as well

## Summary by CodeRabbit

* **New Features**
  - Unified internal TypeScript generation engine while keeping public interfaces compatible.

* **Chores**
  - Deferred TypeScript import to reduce memory usage during normal operations.
  - Bumped package version to 26.1.0-beta.2.
  - Updated changelog and added a contributor entry.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/pullfrog.yml
+++ b/.github/workflows/pullfrog.yml
@@ -1,0 +1,42 @@
+# PULLFROG ACTION — DO NOT EDIT EXCEPT WHERE INDICATED
+name: Pullfrog
+on:
+  workflow_dispatch:
+    inputs:
+      prompt:
+        type: string
+        description: Agent prompt
+  workflow_call:
+    inputs:
+      prompt:
+        description: Agent prompt
+        type: string
+
+permissions:
+  id-token: write 
+  contents: read
+
+jobs:
+  pullfrog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      # Optionally, setup your repo here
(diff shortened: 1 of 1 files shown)


express-zod-api/src/zts.ts
NicolasMahe: @RobinTail typescript is still imported here. Related to https://github.com/RobinTail/express-zod-api/discussions/3126#discussioncomment-15326571

express-zod-api/src/integration-base.ts
```diff
@@ -1,51 +1,16 @@
...
-  literally,
-} from "./typescript-api";
+import { propOf, Typeable, TypescriptAPI } from "./typescript-api";
 
 type IOKind = "input" | "response" | ResponseVariant | "encoded";
 type SSEShape = ReturnType<typeof makeEventSchema>["shape"];
 type Store = Record<IOKind, ts.TypeNode>;
 
-export abstract class IntegrationBase {
+export abstract class IntegrationBase extends TypescriptAPI {
```
RobinTail: @pullfrog instead of extending, please make a protected property `api` assigned with an instance of `TypescriptAPI` class and change the implementation of the `Integration` and `IntegrationBase` classes accordingly

RobinTail (commented):
🏁 ready

PR #3345, opened 2026-04-24
fix: rm redundant `pullResponseExamples` call from `Endpoint::outputSchema`

RobinTail:
Performance regression since v24.0.0 - the `outputSchema` getter was calling the expensive `pullResponseExamples()` operation on first access (R.once protected), even though examples are only needed for OpenAPI documentation generation (getResponses).

This fixes the issue where simply accessing the output schema for runtime validation was unnecessarily triggering example generation.

first introduced by #2666 
merged by #2546 
found this during my work on #3344 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request performance by deferring output schema example processing; examples are now generated only when needed for documentation/OpenAPI workflows, not during initial request handling.

* **Documentation**
  * Added changelog entry v27.2.4 documenting the change in example processing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Fixed performance regression since v24.0.0:
   - Removed unnecessary processing of output schema examples;
   - Those used to be processed when the first API request occurred on endpoint;
-  - Examples are only required for OpenAPI generators, not for runtime validation.
+  - Examples are only required for Documentation generator, not for runtime validation.
 
 ### v27.2.3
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Version 27
 
+### v27.2.4
+
+- Fixed performance regression since v24.4.0:
```
RobinTail:
```suggestion
- Fixed performance regression since v24.0.0:
```

PR #3351, opened 2026-04-25
ref(mem,perf): Turning well-known headers into a `Set`

RobinTail:
This should solve two problems:
1. Unnecessary memory consumption (2k) by the large array that is statically imported while only used by Documentation (not API runtime);
2. `Array::includes()` is O(n) while `Set::has` is O(1): in current situation it could give about 5.5x performance improvement in worst scenario (finding last entry)

The solution, however, requires a lazy creation of the `Set` because it uses even more memory, so:
- ~~`R.once()` is involved, which is also reducing the expected performance boost down to 3.58x~~,
- module cache with nullish coalescing assignment used giving 4.8x boost (better than R.once)

## Summary by CodeRabbit

* **Performance Improvements**
  * Header lookup now uses a memoized set for faster, lower-memory checks.

* **Tests**
  * Added benchmarks measuring header lookup performance.
  * Added unit tests validating the memoized header collection.
  * Removed an obsolete queue-operations benchmark.

* **Chores**
  * Updated generated tooling output to use the new header lookup approach.

* **Changelog**
  * Recorded release v27.2.5 noting memory and lookup improvements.


RobinTail:
@pullfrog evaluate the approach thoughtfully: explain the impact on memory consumption and performance both for API runtime and when running Documentation generator.

RobinTail:
this may accompany another Set lookup in #3344

PR #3354, opened 2026-04-29
fix(test): Handling the external validator malfunction

RobinTail:
making test optional when the external tool fails

## Summary by CodeRabbit

* **Tests**
  * Updated tests with explicit timeouts and automatic abort for long-running requests.
  * Improved error handling so assertions run only on successful responses and failures are asserted cleanly.
  * Moved shared test data out of individual test bodies for reuse and clarity.


RobinTail:
should be more complicated, but the tool seems to be operational now

PR #3407, opened 2026-05-20
exp: intake contract for ResultHandler

RobinTail:
Exploring the DX trade-off between the requirement for ResultHandler to handle any object-based `output` coming from Endpoints, or giving an ability to enforce type-only contract between ResultHandler and Endpoint through EndpointsFactory.

The POC looks promising except the `EndpointsFactory::buildVoid()` method that stands out of the concept.

The risk of possible temptation to abuse the type-only contract and remove runtime checks completely may compromise the stability of final products. Therefore, it's still under consideration.


RobinTail:
the idea kinda operational, but the tradeoff is too risky

PR #3413, opened 2026-05-23
exp: Simpler `defaultResultHandler` and status-encoded Client types

RobinTail:
The idea is to change the payload of `defaultResultHandler` in order to make it smaller and simpler by removing the `status: success | error` discriminator as well as `data | error` wrapper. The status is already conveyed by the HTTP status code, so it is redundant.

That, however, requires to discriminate by the HTTP status on consumer side. Luckily, we do already have an entity called `EncodedResponse` (introduced as a result of #2219 discussion). With minor changes it can be used as a tuple-driven discriminator for the returns of `Client::provide()`.

The whole change is breaking both for developers and the consumers of their APIs. Therefore a comprehensive migration path is required.

My current concern is that the connection between the response types and the status codes is too strict: when receiving an undeclared code, the payload type would be `never`. I'd like to offer something as a fallback instead.


example/index.spec.ts
```diff
@@ -127,9 +124,9 @@ describe("Example", async () => {
       );
       expect(response.status).toBe(200);
       const json = (await response.json()) as {
-        data?: { users?: Array<{ role: string }> };
+        users?: Array<{ role: string }>;
       };
-      const users = json.data?.users;
+      const users = json.users;
```
RobinTail: consider deconstruction here

example/example.client.ts
```diff
@@ -816,9 +542,13 @@ export class Client<T> {
...
     ctx?: T,
-  ): Promise<Response[K]> {
+  ): Promise<EncodedResponse[K]> {
     const [method, path] = parseRequest(request);
-    return this.implementation(method, ...substitute(path, params), ctx);
+    return this.implementation(
+      method,
+      ...substitute(path, params),
+      ctx,
+    ) as Promise<EncodedResponse[K]>;
```
RobinTail: check if this could be avoided

RobinTail: no

RobinTail:
When deconstructung the tuple `const [status, response] = Client.provide()` and narrowing the `status` using a static `if (Client::isSuccess(status))` it does not narrow the `response` within the block. Only the explicit `===` work.

I wanna try a different approach. Though, the overall direction remains interesting.

PR #3447, opened 2026-06-07
exp: OpenAPI component individual descriptions

RobinTail:
Exploring better ways than `descriptions` (record of callbacks) of Documentation argument.

At least for responses I am aiming to provide a way to express the meaning of the payload instead of generic way to encode HTTP properties.

### DIFF
--- a/example/example.documentation.yaml
+++ b/example/example.documentation.yaml
@@ -195,7 +195,7 @@ paths:
           APIKEY_2: []
       responses:
         "200":
-          description: PATCH /v1/user/:id Positive response
+          description: Successful response
           content:
             application/json:
               schema:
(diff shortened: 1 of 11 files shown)


express-zod-api/src/result-handler.ts
```diff
@@ -130,6 +131,7 @@ export const defaultResultHandler = new ResultHandler({
     const examples = getExamples(output); // pulling down:
     if (examples.length) {
       globalRegistry.add(responseSchema, {
+        description: output.description,
```
RobinTail: questionable move. Should probably be just "Successful response", generic

example/example.documentation.yaml
```diff
@@ -74,6 +74,7 @@ paths:
                   - status
                   - error
                 additionalProperties: false
+                description: Error response
```
RobinTail: should we pull it as we do for examples?

RobinTail:
technically it works, but semantically it's not correct

PR #3448, opened 2026-06-07
exp: Wildcard status codes

RobinTail:
The feature of OpenAPI
https://swagger.io/specification/#responses-object


RobinTail:
this should be not the DX, but the generator's work

PR #3492, opened 2026-06-27
fix(mem): Destroy and remove failed sockets by graceful monitoring

RobinTail:
This is supposed to handle broken sockets that emit `error` but do not emit `close`, so that the `Set` won't grow.
This is an edge case.
Extracted from #3490 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved socket cleanup during graceful shutdown so tracked connections are removed correctly even when a socket errors.
  * Ensured errored sockets are destroyed and no longer linger in the active socket set.
* **Tests**
  * Added coverage for socket error handling during shutdown to verify cleanup and shutdown complete successfully.


RobinTail:
this is more like an edge case, since Node.js docs says:

> Emitted when an error occurs. The 'close' event will be called directly following this event.

PR #3528, opened 2026-07-13
exp: Reducing schema flattening

RobinTail:
This is a heavy operation that could be probably simplified using hooks.


RobinTail:
no longer an option since checking more than just path params #3596

PR #3588, opened 2026-08-01
fix(DNRY): Reduce flattening in Documentation

RobinTail:
I noticed that some operations could be executed once and reused.

## Summary by CodeRabbit

* **Documentation**
  * Improved generated API documentation for request parameters and request bodies.
  * Request fields are represented more accurately across parameters and body content.
  * Enhanced handling of required fields, nested object structures, fallback examples, and raw request bodies.
  * Path parameters are now excluded from request-body schemas and examples.

* **Tests**
  * Expanded documentation coverage for request parameter and body representations, including parameter exclusion scenarios.


express-zod-api/tests/documentation-helpers.spec.ts
RobinTail, line 502: @pullfrog The changed fixture only verifies that ez.raw() sets required. Add a depictBody case with paramNames and flat.examples to verify that fallback examples omit parameter fields. Add an integration case for documentation.ts to verify that the prepared jsonSchema excludes parameter fields before body depiction.

PR #3593, opened 2026-08-03
br(deps,CI): Restore entirely private in DTS, drop TS5 support

RobinTail:
~~I'm not sure why, but~~ having `#private` property in the class declaration seems to be no longer an issue (no errors).
Logs are not retained in #3272 

To be sure, this requires to drop TypeScript 5 support first, so keeping it for v30

The issue was in `issue-952` test: `Property '#private' of exported anonymous class type may not be private or protected.`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved `#private` markers in generated declaration files.
  * Continued cleanup of region comments and duplicate blank lines before formatting.


RobinTail:
@pullfrog could you figure out the reason why `#private` caused the issue in `issue-952` test as of `2b1e58cf` state, but now it does not?

PR #3634, opened 2026-08-15
fix(perf): Using `Array.from()` built-in mapper

RobinTail:
## Summary by CodeRabbit

* **Refactor**
  * Improved internal array processing performance without changing generated output or externally visible behavior.

* **Chores**
  * Added linting guidance to identify less efficient array transformation patterns.


RobinTail:
```
  · Array.from().map()  24,497,460.48  0.0000  0.1416  0.0000  0.0000  0.0001  0.0001  0.0002  ±0.23%  12248731
   · Array.from(map)      7,022,036.90  0.0000  0.1617  0.0001  0.0002  0.0002  0.0003  0.0005  ±0.12%   3511019
```
bench displayed the opposite result

PR #3662, opened 2026-08-26
fix: document cookie-only inputs and QUERY query params

### LINKED ISSUE #3660: Cookie-sourced input fields are omitted or documented as query unless declared in CookieSecurity
# Description

`makeParamLocator` only assigns `in: cookie` to names listed in a `CookieSecurity` declaration. Every other property that is actually read from `request.cookies` (because `"cookies"` / `"signedCookies"` is in `inputSources`) falls through to the query fallback, or to nowhere.

Two outcomes, depending on whether `"query"` is also enabled:

1. **`inputSources` includes cookies but not query.** The field is accepted from the cookie at runtime and rejected from the query string, yet the generated operation has **no parameter for it at all**. A client generated from that document cannot send a valid request. This is the same category as #3577 / #3578: a contract the endpoint cannot fulfil as documented (here, because the required input is missing from the document).
2. **`inputSources` includes both query and cookies** (the README example). The field is documented as `in: query`. Runtime still accepts the cookie, and with cookies listed after query the cookie value overwrites the query value. The document therefore points clients at the lower-priority source and never mentions the cookie.

Adding `security: { type: "cookie", name: "session" }` makes the first name appear as `in: cookie`, which is why the README presents CookieSecurity as improving Documentation. Non-security cookie fields have no such escape hatch: `isHeader` can be customized for headers, but cookies have no equivalent.

Routing itself is unaffected. Only the generated document is wrong.

The same locator is also why HTTP `QUERY` never emits `in: query` parameters (`method !== "query"` skips the query fallback and leaves those fields in the form body), even though the default input sources for QUERY are `["query", "body", "params"]` and the URL query string is merged at runtime. That case is milder, because putting the field in the body still works, but it comes from the same `getLocation` path.

## Expected

When `"cookies"` or `"signedCookies"` is among the method's input sources, properties that are not path/header parameters should be eligible for `in: cookie` without requiring a matching `CookieSecurity` name. Security can still improve the `securitySchemes` section; it should not be the only way to get the correct parameter location.

For QUERY, properties that are not path parameters should still be able to appear as `in: query` when `"query"` is among the input sources, consistent with the default sources and with GET.

## Reproduction

```js
import { z } from "zod";
import {
  createConfig,
  createServer,
  defaultEndpointsFactory,
} from "express-zod-api";
import { Documentation } from "express-zod-api/documentation";

const port = 8100;

const config = createConfig({
  http: { listen: port },
  cors: false,
  startupLogo: false,
  logger: { level: "warn" },
  cookies: true,
  // cookies enabled; query deliberately omitted
  inputSources: { get: ["params", "cookies"] },
});

const ep = defaultEndpointsFactory.build({
  method: "get",
  input: z.object({ session: z.string() }),
  output: z.object({ received: z.string() }),
  handler: async ({ input }) => ({ received: JSON.stringify(input) }),
});

const routing = { "get /c": ep };

const doc = new Documentation({
  routing,
  config,
  info: { title: "T", version: "1.0.0" },
}).getSpec();

console.log("parameters:", JSON.stringify(doc.paths["/c"].get.parameters));

await createServer(config, routing);

for (const [label, url, headers] of [
  ["as query (as a client might guess)", `http://localhost:${port}/c?session=s`, {}],
  ["as cookie (actual source)", `http://localhost:${port}/c`, { cookie: "session=s" }],
]) {
  const res = await fetch(url, { headers });
  console.log(label, res.status, await res.text());
}

process.exit(0);
```

Output:

```
parameters: undefined
as query (as a client might guess) 400 {"status":"error","error":{"message":"session: Invalid input: expected string, received undefined"}}
as cookie (actual source) 200 {"status":"success","data":{"received":"{\"session\":\"s\"}"}}
```

The document exposes no way to supply `session`. The server requires it from the cookie.

With the README-shaped sources `{ get: ["query", "params", "cookies"] }` and the same schema, the document instead emits `session` as `in: query`. Runtime accepts both, and the cookie wins when both are present:

```
parameters: [{"name":"session","in":"query",...},{"name":"q","in":"query",...}]
as query          200 ..."session":"s"...
session as cookie 200 ..."session":"s"...
both              200 ..."session":"from-cookie"...   // cookie overwrote query
```

## Context

- Node.js version: 24.14.0
- `express-zod-api` version: 29.3.4
- `zod` version: 4.4.3

BetterAndBetterII:
## Summary

`makeParamLocator` only emitted `in: cookie` for names listed in `CookieSecurity`. When `inputSources` included `cookies`/`signedCookies` but not `query`, remaining fields were omitted from the OpenAPI document even though runtime accepted them from cookies.

The same locator skipped the query fallback for HTTP `QUERY` (`method !== "query"`), so QUERY fields landed in the form body instead of `in: query` despite default sources including `query`.

## Changes

- When cookies are enabled and query is not, remaining fields are documented as `in: cookie` without requiring CookieSecurity.
- CookieSecurity still wins when both query and cookies are enabled.
- QUERY methods can emit `in: query` when query is among inputSources.

## Test plan

- [x] `vitest run tests/documentation-helpers.spec.ts tests/documentation.spec.ts` (151 passed)
- [x] Unit coverage for cookie-only sources and QUERY query params
- [x] Documentation integration expectations for both cases

Fixes #3660

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API documentation for cookie-sourced parameters, including unsigned and signed cookies.
  * Query parameters are now correctly identified for query-based endpoints when query input is enabled.
  * Cookie parameters no longer require additional cookie security configuration to be documented accurately.

* **Tests**
  * Added coverage for cookie and query parameter documentation scenarios.

### DIFF
--- a/express-zod-api/src/documentation-helpers.ts
+++ b/express-zod-api/src/documentation-helpers.ts
@@ -287,7 +287,8 @@ export const makeParamLocator = ({
       (isHeader?.(name, method, path) ?? defaultIsHeader(name, securityHeaders))
     )
       return "header";
-    if (isQueryEnabled && method !== "query") return "query";
+    if (isQueryEnabled) return "query";
+    if (areCookiesEnabled) return "cookie";
   };
   return { pathParams, getLocation, isQueryEnabled };
 };
--- a/express-zod-api/tests/__snapshots__/documentation.spec.ts.snap
+++ b/express-zod-api/tests/__snapshots__/documentation.spec.ts.snap
@@ -949,26 +949,29 @@ paths:
   /v1/getSomething:
     query:
       operationId: QueryV1GetSomething
+      parameters:
+        - name: intersection
+          in: query
+          required: true
+          description: QUERY /v1/getSomething Parameter
+          schema:
+            type: object
+            properties:
+              one:
+                type: string
+              two:
+                type: string
+            required:
+              - one
+              - two
       requestBody:
         description: QUERY /v1/getSomething Request body
         content:
           application/x-www-form-urlencoded:
             schema:
               type: object
-              properties:
-                intersection:
-                  type: object
-                  properties:
-                    one:
-                      type: string
-                    two:
-                      type: string
-                  required:
-                    - one
-                    - two
-              required:
-                - intersection
-        required: true
+              properties: {}
+              required: []
       responses:
         "200":
           description: QUERY /v1/getSomething Positive response
--- a/express-zod-api/tests/documentation-helpers.spec.ts
+++ b/express-zod-api/tests/documentation-helpers.spec.ts
@@ -446,6 +446,32 @@ describe("Documentation helpers", () => {
       expect(getLocation("test")).toBe("query");
       expect(getLocation("session")).toBe("cookie");
     });
+
+    test.each(["cookies", "signedCookies"] as const)(
+      "Issue #3660: should depict %s as cookie params without CookieSecurity",
+      (source) => {
+        const { pathParams, isQueryEnabled, getLocation } = makeParamLocator({
+          ...requestCtx,
+          inputSources: ["params", source],
+        });
+        expect(pathParams).toEqual(new Set(["id"]));
+        expect(isQueryEnabled).toBe(false);
+        expect(getLocation("id")).toBe("path");
+        expect(getLocation("session")).toBe("cookie");
+      },
+    );
+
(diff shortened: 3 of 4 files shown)


RobinTail:
@BetterAndBetterII , thank you for the contribution.
Though this PR is not gonna be merged, because the original problem requires a little more comprehensive solution, some of your test cases are useful and will be reused as part of #3682

PR #3666, opened 2026-08-31
feat(perf)!: Using compiled schemas

RobinTail:
- Min zod version 4.5.4;
- Enabling sync parsing attempt first (for compiled schema benefits);
- Wrapping I/O schemas with `z.compile()`, but ensuring that we compile once to reuse compiled instances where same schema reused between endpoints;
- The unstable traverse problem for compiled schemas is addressed in #3680 

### Performance benchmark

- building: becomes ~2.38x slower (compiling once at startup)
- parsing: becomes ~1.62x faster (each API request)

## Summary by CodeRabbit

* **New Features**
  * Added support for Zod versions `^4.5.4`.
  * Added compiled I/O schema support to improve endpoint runtime performance.
  * Preserved shared output schemas as a single deduplicated component in generated OpenAPI documentation.

* **Bug Fixes**
  * Improved schema reuse and compilation consistency across endpoints and middleware.


RobinTail:
compiled schemas seem to have a slightly different order in `toJSONSchema()`

RobinTail:
I didn't get it from the release notes but it seems that it won't work with `parseAsync()` method which is exactly one used in Endpoint runtime

RobinTail:
I have an idea: we should try regular `.parse()` first. Unless it throws `$ZodAsyncError`. I used the `parseAsync()` in Endpoint for versatility, but it should now become more sophisticated. #3672

RobinTail:
> All three runtime validation sites — endpoint.ts:213 (output), endpoint.ts:261 (input), and middleware.ts:122 — parse with .parseAsync()

This is now fixed — we now attempt to parse sync first (#3673), @pullfrog .
Please review again.

express-zod-api/src/endpoints-factory.ts
```diff
@@ -214,7 +214,7 @@ export class EndpointsFactory<
     return new Endpoint({
       ...rest,
       middlewares,
-      outputSchema,
+      outputSchema: z.compile(output),
```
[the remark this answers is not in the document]
RobinTail: any suggestions in this regard? should we maintain a WeakMap of compiled schemas? Or should we delegate the schema compilation to the framework users? btw, is there any property of the schema that could tell if the schema is compiled or not? @pullfrog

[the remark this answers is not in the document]
RobinTail: this concern was addressed by [c32f53b](https://github.com/RobinTail/express-zod-api/pull/3666/commits/c32f53b8beecd4579ad2cfe97680c991378f675d)

express-zod-api/src/common-helpers.ts
```diff
@@ -13,12 +13,29 @@ import {
...
 
+/**
+ * @desc Compiles the schema once and caches the compiled clone per schema instance.
+ * @desc Keeps the identity of schemas shared between Endpoints and Middlewares so that they are not recompiled on
+ *       each construction and the component deduplication in the Documentation keeps working.
+ * @desc Schemas already compiled by z.compile() or the zod global shim are passed through as is.
+ * */
+const compiledSchemas = new WeakMap<object, object>();
+export const compileOnce = <T extends z.ZodType>(schema: T): T => {
+  if ((schema._zod.run as { __originalRun?: unknown }).__originalRun)
```
RobinTail: @pullfrog should we add alternative checks such as `validator` and `fallbackRun` on the `._zod.bag` to make it more future-proof to potential changes in the external library?

express-zod-api/src/common-helpers.ts
```diff
@@ -13,12 +13,34 @@ import {
...
+ * @desc Schemas already compiled by z.compile() or the zod global shim are passed through as is, detected by the
+ *       markers that the external library sets on the instance internals.
+ * */
+const compiledSchemas = new WeakMap<object, object>();
+export const compileOnce = <T extends z.ZodType>(schema: T): T => {
+  if (
+    (schema._zod.run as { __originalRun?: unknown }).__originalRun ||
+    schema._zod.bag.validator ||
+    schema._zod.bag.fallbackRun
+  )
```
RobinTail: @pullfrog I suggest to extract this into a new `isCompiled()` helper and test it independently, so we won't test every case for `compileOnce`.

express-zod-api/src/common-helpers.ts
```diff
@@ -13,12 +13,40 @@ import {
...
+    schema._zod.bag.fallbackRun,
+  );
+
+/**
+ * @desc Compiles the schema once and caches the compiled clone per schema instance.
+ * @desc Keeps the identity of schemas shared between Endpoints and Middlewares so that they are not recompiled on
+ *       each construction and the component deduplication in the Documentation keeps working.
+ * @desc Schemas already compiled (see isCompiled()) are passed through as is.
+ * */
+const compiledSchemas = new WeakMap<object, object>();
```
RobinTail: if the schema identity is shared between endpoints, it might mean that it's assigned to a const. Then, does it mean that it will remain forever in the `WarkMap`, @pullfrog ?

RobinTail: @pullfrog in the `example/endpoints/list-users-paginated.ts` we have `paginatedUsers` entity and its `.output` schema is used in the file as the output schema. I tested by exposing `compiledSchemas` publicly, imported it in the `example/index.ts` as well as `paginatedUsers` and even after the server started `console.log(hasCompiled(paginatedUsers.output));` says `true`. So my concern is valid. A schema that persists in a `const` also persists in the `WeakMap` and I don't think GC will eventually remove it from there because the `const` reference remains. I'm not sure if it's a memory leak, because we do use the compiled schema from that WeakMap. But we don't use the non-compiled one.

express-zod-api/tests/env.spec.ts
```diff
@@ -3,6 +3,7 @@ import * as R from "ramda";
 import { z } from "zod";
 import { createRequire } from "node:module";
 import { METHODS } from "node:http";
+import { expect } from "vitest";
```
RobinTail:
```suggestion
```

express-zod-api/src/io-schema.ts
```diff
@@ -42,6 +43,6 @@ export const makeFinalInputSchema = <
   factorySchema: FIN,
   buildSchema: BIN,
 ) =>
-  (factorySchema
-    ? factorySchema.and(buildSchema)
-    : buildSchema) as FinalInputSchema<FIN, BIN>;
+  compileOnce(
+    factorySchema ? factorySchema.and(buildSchema) : buildSchema,
+  ) as FinalInputSchema<FIN, BIN>;
```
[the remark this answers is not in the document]
RobinTail: @pullfrog implement such a bench

RobinTail (commented):
🏁

